PR #91, opened 2024-06-05
#110 Change auth to session. 

martineiber:
Part of https://github.com/pimcore/studio-ui-bundle/issues/110

## Additional info

Update your `security.yaml` in your project:

```yaml
security:
    firewalls: 
        pimcore_studio: '%pimcore_studio_backend.firewall_settings%'
    access_control:
        - { path: ^/studio/api/(docs|docs.json|translations)$, roles: PUBLIC_ACCESS }
        - { path: ^/studio, roles: ROLE_PIMCORE_USER }
```


fashxp:
please add firewall info to readme file too

mattamon:
Repo is now public, comments can be seen by everyone :) We should stick to english

fashxp:
> Repo is now public, comments can be seen by everyone :) We should stick to english

ups, you're right :smile:

config/authorization.yaml
```diff
@@ -13,5 +13,6 @@ services:
     tags: [ 'controller.service_arguments' ]
 
 
-  Pimcore\Bundle\StudioBackendBundle\Authorization\Service\TokenServiceInterface:
-    class: Pimcore\Bundle\StudioBackendBundle\Authorization\Service\TokenService
+  Pimcore\Bundle\StudioBackendBundle\Authorization\EventListener\LogoutListener:
```
mattamon: Should we maybe move it to a subscriber namespace since it is not really a listener and in the src directory I already created a `EventSubscriber` folder? :) Wdyt?

src/Authorization/Controller/LoginController.php
```diff
@@ -0,0 +1,57 @@
...
+        operationId: 'login',
+        summary: 'Login with user credentials and get access token',
+        tags: [Tags::Authorization->name]
+    )]
+    #[CredentialsRequestBody]
+    #[SuccessResponse(
+        description: 'Login successful',
+        content: new JsonContent(ref: LoginSuccess::class)
+    )]
+    #[InvalidCredentials]
```
mattamon: #[DefaultResponses()]
    
    We should add the default responses here too in case anything else goes wrong or can we be sure that this will never happen?

src/Authorization/EventListener/LogoutListener.php
```diff
@@ -0,0 +1,39 @@
...
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            LogoutEvent::class => 'onLogout',
+        ];
+    }
+
+    public function onLogout(LogoutEvent $event): void
+    {
+        $event->setResponse(new JsonResponse([]));
```
mattamon: Does it have to be a JsonResponse with an empty array or can it be just a standard http response?

src/DependencyInjection/PimcoreStudioBackendExtension.php
```diff
@@ -72,13 +74,33 @@ public function load(array $configs, ContainerBuilder $container): void
...
+
+    public function prepend(ContainerBuilder $container): void
+    {
+        if (!$container->hasParameter('pimcore_studio_backend.firewall_settings')) {
+            $containerConfig = ConfigurationHelper::getConfigNodeFromSymfonyTree($container, 'pimcore_studio_backend');
+            $container->setParameter('pimcore_studio_backend.firewall_settings', $containerConfig['security_firewall']);
+        }
+    }
+
+    private function checkValidOpenApiScanPaths(array $config): void
```
mattamon, line 96: Please add the throw tags in the doc block :)

src/Security/Service/SecurityService.php
```diff
@@ -75,23 +72,16 @@ public function checkAuthToken(string $token): bool
...
      */
     public function getCurrentUser(): UserInterface
     {
-        $entry = $this->tmpStoreResolver->get($this->tokenService->getCurrentToken());
-
-        if($entry === null || !is_array($entry->getData()) || !isset($entry->getData()['username'])) {
-            throw new NotAuthorizedException();
+        $pimcoreUser = $this->authenticationResolver->authenticateSession();
+        if (!$pimcoreUser instanceof User) {
+            throw new UserNotFoundException('No pimcore user found');
```
mattamon, line 50: Should we really switch to a user not found exception instead of a NotAuthorized?

mattamon (changes requested):
Works like a charm :)

src/Security/Service/SecurityService.php
martineiber: Since the Token check was removed, we do not really check for authentication here. We just grab the Authenticated user. So that's my i would leave this Exception.

mattamon: But if you try to grab the user the only reason why you would not get it, is because you are not authorized right? :)

martineiber: IMHO then you should not be able to enter the Controller at all. And this is done by the Access Control. 

And since the method says getCurrentUser. I think it makes more sense to throw a UserNotFoundException.

mattamon: Agreed! :)

mattamon (approved):
LGTM 🦭

martineiber:
We will keep this PR open until the FE Part is ready to merge. Otherwise we will break the FE.

PR #93, opened 2024-06-06
Asset Image sidebar - add endpoints for asset and image download

lukmzig:
Related to https://github.com/pimcore/studio-ui-bundle/issues/214

## Additional info
Added endpoints for 
- original asset download
- image download by format
- image download by configuration
- image download by thumbnail name


src/Asset/Controller/DownloadController.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * @throws AccessDeniedException
+     * @throws ElementStreamResourceNotFoundException
+     * @throws ElementNotFoundException
+     * @throws InvalidElementTypeException
+     * @throws UserNotFoundException
+     */
+    #[Route('/assets/{id}/download', name: 'pimcore_studio_api_download_asset', methods: ['GET'])]
+    //#[IsGranted('STUDIO_API')]
+    //#[IsGranted(UserPermissions::ASSETS->value)]
```
martineiber: IsGranted Should be used?

src/Exception/ElementStreamResourceNotFoundException.php
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @internal
+ */
+final class ElementStreamResourceNotFoundException extends AbstractApiException
+{
+    public function __construct(int $id, string $type = 'Element')
+    {
+        parent::__construct(
+            404,
```
martineiber: Here the same.. try to use the HttpCode Enum :)

src/Util/Traits/ConsoleExecutableTrait.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+trait ConsoleExecutableTrait
+{
+    /**
+     * @throws ConsoleDependencyMissingException
+     */
+    private function getExecutable(string $executable, string $module = 'Pimcore'): string
+    {
+        try {
+            $consoleExecutable = Console::getExecutable($executable);
```
martineiber, line 34: Can we use the static resolver here in the trait? Or will it be just to much overhead?

martineiber (changes requested):
Some minor changes.

And here i'm also missing some Unit Test ;)

src/Util/Traits/ConsoleExecutableTrait.php
lukmzig: Not sure if this would make much sense?
We would need then pass the resolver to the trait and we cannot write usable unit tests here anyway

martineiber: Then i would say we leave it like it is :)

src/Asset/Controller/DownloadController.php
lukmzig: Thank you, I missed this one

martineiber (approved):
LGTM

PR #95, opened 2024-06-06
[Task] Update schema require

### LINKED ISSUE #88: [Task] Check schema required fields
By default every property of a schema is nullable, check if we have to add the required fields in the schema definition

mattamon:
Resolves #88


mattamon:
Closed in favor of #111

PR #99, opened 2024-06-07
#179 List all Users. 

martineiber:
Part of https://github.com/pimcore/studio-ui-bundle/issues/179


src/User/Hydrator/UserTreeNodeHydrator.php
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * @internal
+ */
+final class UserTreeNodeHydrator implements UserTreeNodeHydratorInterface
+{
+    public function hydrate(User|User\Folder $user): UserTreeNode
+    {
+        $hasChildren = false;
+        if ($user instanceof User\Folder) {
```
mattamon: Can we fully import that folder?

src/User/Controller/CollectionController.php
```diff
@@ -0,0 +1,74 @@
...
+    #[ParentIdParameter(
+        description: 'Filter users by parent id.',
+        required: true
+    )]
+    #[SuccessResponse(
+        description: 'Collection of users including folders for the given parent id.',
+        content: new CollectionJson(new GenericCollection(UserTreeNode::class))
+    )]
+    #[DefaultResponses([
+        HttpResponseCodes::NOT_FOUND
```
mattamon, line 67: When would the collection controller return a not found reponse?

mattamon (changes requested):
Rest LGTM :)

src/User/Controller/CollectionController.php
martineiber: When the ParentId is Blank or negative Symfony returns a 404.

src/User/Hydrator/UserTreeNodeHydrator.php
martineiber: Fixed in [be905a8](https://github.com/pimcore/studio-backend-bundle/pull/99/commits/be905a865c71b53a18d7b36cb8adef799057a4e7)

src/User/Controller/CollectionController.php
mattamon: But then we need to add this to the default responses because this can happen at any route then, no?

martineiber: Which makes sense since the UserList for a Negative Parent id does not exist.

martineiber: If there is no check for any Parameter this can not happen, or if there is at all no parameter. So i would say no, this is not the Default.

mattamon (approved):
LGTM

PR #118, opened 2024-06-12
#179 Endpoint for deleting users and user-folders

martineiber:
Part of https://github.com/pimcore/studio-ui-bundle/issues/179


src/Exception/ForbiddenException.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Pimcore\Bundle\StudioBackendBundle\Exception;
+
+/**
+ * @internal
+ */
+final class ForbiddenException extends AbstractApiException
+{
+    public function __construct(string $message = 'Access Denied')
+    {
+        parent::__construct(403, $message);
```
mattamon: We should use the HttpResponseCodes enum here. I already changed it for the openapi responses, maybe we should use them in the exceptions too? WDYT?

src/User/Controller/DeleteUserController.php
```diff
@@ -0,0 +1,67 @@
...
+
+
+    #[Route('/user/{id}', name: 'pimcore_studio_api_user_delete', methods: ['DELETE'])]
+    #[Delete(
+        path: self::API_PATH . '/user/{id}',
+        operationId: 'deleteUser',
+        summary: 'Delete a specific user.',
+        tags: [Tags::User->value]
+    )]
+    #[IsGranted(UserPermissions::USER_MANAGEMENT->value)]
```
mattamon: I would move it under the route, so the structure is more like first symfony attributes, then openapi attributes

src/User/Controller/DeleteUserFolderController.php
```diff
@@ -0,0 +1,66 @@
...
+
+
+    #[Route('/user/folder/{id}', name: 'pimcore_studio_api_user_folder_delete', methods: ['DELETE'])]
+    #[Delete(
+        path: self::API_PATH . '/user/folder/{id}',
+        operationId: 'deleteUserFolder',
+        summary: 'Delete a specific user folder with all users in this folder.',
+        tags: [Tags::User->value]
+    )]
+    #[IsGranted(UserPermissions::PIMCORE_ADMIN->value)]
```
mattamon: Same here

src/User/Controller/DeleteUserFolderController.php
```diff
@@ -0,0 +1,66 @@
...
+    #[SuccessResponse]
+    #[IdParameter(type: 'user-folder')]
+    #[DefaultResponses([
+        HttpResponseCodes::NOT_FOUND
+    ])]
+    public function deleteUserFolder(int $id): Response
+    {
+        $this->userFolderService->deleteUserFolderById($id);
+        return new Response();
```
mattamon, line 70: For other delete routes we return the deleted id, instead of only a success response.
Should we maybe make this consistent too?  
Either only a success response or always the id, I have no preference.

src/User/Controller/DeleteUserFolderController.php
```diff
@@ -0,0 +1,66 @@
...
+use Pimcore\Bundle\StudioBackendBundle\Util\Constants\HttpResponseCodes;
+use Pimcore\Bundle\StudioBackendBundle\Util\Constants\UserPermissions;
+use Pimcore\Bundle\StudioBackendBundle\Util\Traits\PaginatedResponseTrait;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Attribute\Route;
+use Symfony\Component\Security\Http\Attribute\IsGranted;
+use Symfony\Component\Serializer\SerializerInterface;
+
+/**
+ * @internal
```
mattamon, line 38: The Controller should also have all the throw tags

src/Exception/ForbiddenException.php
```diff
@@ -0,0 +1,28 @@
...
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Exception;
+
+/**
+ * @internal
+ */
+final class ForbiddenException extends AbstractApiException
```
mattamon, line 22: Should we maybe call it AccessDeniedException?

mattamon (changes requested):
Rest looks already good to me

src/User/Controller/DeleteUserFolderController.php
martineiber: We will go with an empty 200 Response

mattamon: 200 Response is enough No ID needed. Unify PR already open. https://github.com/pimcore/studio-backend-bundle/pull/122

mattamon (approved):
LGTM

PR #145, opened 2024-06-17
Use GDI Permission check for dependencies

### LINKED ISSUE #66: [Dependency] Permissions
Currently there is a permissions check for the requested element only. 
But the according dependencies have to be checked for the `listing` permission too, see existing implementation: https://github.com/pimcore/pimcore/blob/e9c9e2d43e56502f000a38c59eb7f92c6029c347/models/Element/Service.php#L192

The idea is to have an endpoint in `generic-data-index-bundle` that can check permissions for more than one object, thus providing a fast way to check permissions on a set of elements.

mattamon:
Resolves #66

### DIFF
--- a/src/Dependency/Hydrator/DependencyHydrator.php
+++ b/src/Dependency/Hydrator/DependencyHydrator.php
@@ -36,6 +36,9 @@ final readonly class DependencyHydrator implements DependencyHydratorInterface
 
     public function hydrate(ElementSearchResultItemInterface $dependency): ?Dependency
     {
+        // isPublished does not exist in the ElementSearchResultItemInterface
+        // unfortunately there is no other interface for is published
+        // documents and objects have the isPublished method
         return new Dependency(
             $dependency->getId(),
             $dependency->getFullPath(),
--- a/src/Dependency/Hydrator/DependencyHydrator.php
+++ b/src/Dependency/Hydrator/DependencyHydrator.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
 
 namespace Pimcore\Bundle\StudioBackendBundle\Dependency\Hydrator;
 
+use Pimcore\Bundle\GenericDataIndexBundle\Model\Search\Interfaces\ElementSearchResultItemInterface;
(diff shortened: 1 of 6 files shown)


src/Dependency/Hydrator/DependencyHydrator.php
```diff
@@ -16,53 +16,25 @@
...
-    use ElementProviderTrait;
-
-    public function __construct(
-        private ServiceResolverInterface $serviceResolver,
-    ) {
-
-    }
-
-    public function hydrate(array $dependency): ?Dependency
+    public function hydrate(ElementSearchResultItemInterface $dependency): ?Dependency
```
martineiber:
```suggestion
    public function hydrate(ElementSearchResultItemInterface $dependency): Dependency
```

nullable can be removed?

src/Dependency/Repository/DependencyRepository.php
```diff
@@ -27,53 +37,54 @@
...
+    ): ElementSearchResult
     {
-        return $this->getElement(
-            $this->serviceResolver,
-            $elementType,
-            $elementId
-        )
-            ->getDependencies()
-            ->getRequires();
+        $search = $this->searchProvider->createAssetSearch();
```
martineiber: Just for my information, the data from the SearchProvider is coming from the Search Server, and we treat it as if it were a DB. Is this why it is in a Repository?

martineiber (changes requested):
Just a small change, Rest LGTM

src/Dependency/Repository/DependencyRepository.php
mattamon: Yeah kind of. Maybe the name is not fitting, but then I really need to refactor the whole thing and maybe put everything into the service or move it to some other class. 
Not sure thought I just tried to make it work with the GDI

src/Dependency/Hydrator/DependencyHydrator.php
mattamon: I removed it. Not sure why it was there in the first place. It is like the permissions. Why would we get an empty dependency from the core, or an empty permissions.

martineiber (approved):
LGTM :+1:

PR #150, opened 2024-06-18
Add updater

### LINKED ISSUE #134: Asset update
Related to https://github.com/pimcore/studio-ui-bundle/issues/173#issuecomment-2152284720

The list of fields for the asset update looks like this:
- partentId
- key
- locked
- all asset metadata attributes for the various types
- focal point 
- properties
- data (for text files the text file content)
- scheduler? (we have there a separate endpoint but in addition to the specific scheduler save button the scheduler should also be saved when the global save button get's pressed. do we solve this with 2 calls?)

In addition to that we need to check that the `userModification` gets automatically updated with the authenticated user.

Later when we build the asset upload we need some more attributes, but I guess this will be then a separate endpoint.

Hopefully I found all attributes but at least there should not be many which are missing in my list.

All of them should be built as discussed in a PATCH update fashion (so that only sent fields are updated and not sent fields remain untouched).

mattamon:
Resolves #134 

## Additional info
Adding updater


lukmzig (approved):
🦭

martineiber (approved):
LGTM

PR #151, opened 2024-06-18
Get element ID by path

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/346


src/DependencyInjection/PimcoreStudioBackendExtension.php
```diff
@@ -77,6 +77,7 @@ public function load(array $configs, ContainerBuilder $container): void
         $loader->load('updater.yaml');
         $loader->load('users.yaml');
         $loader->load('versions.yaml');
+        $loader->load('elements.yaml');
```
mattamon: Do you even alphabet 🔤 :D

src/Element/Request/ElementPathParameter.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Pimcore\Bundle\StudioBackendBundle\Element\Request;
+
+/**
+ * @internal
+ */
+final readonly class ElementPathParameter
+{
+    public function __construct(
+        private string $elementPath
```
mattamon: Could we just name it path?
Wdyt? It is a little redundant in the moment. :)

lukmzig: sure, we can rename it to path :)

src/DependencyInjection/PimcoreStudioBackendExtension.php
lukmzig: abcduuve

src/OpenApi/Config/Tags.php
```diff
@@ -79,12 +83,13 @@ enum Tags: string
...
     case DataObjects = 'DataObjects';
     case Dependencies = 'Dependencies';
+    case Elements = 'Elements';
     case Notes = 'Notes';
     case Properties = 'Properties';
     case Schedule = 'Schedule';
     case Settings = 'Settings';
     case Translation = 'Translation';
-    case Thumbnails = 'Thumbnails';
+    case AssetThumbnails = 'Asset Thumbnails';
```
mattamon: 🔡

mattamon (approved):
(>‿◠)✌

PR #153, opened 2024-06-19
Asset download with GEE

### LINKED ISSUE #137: Asset zip download
For the asset zip download in the asset grid we need to implement an approach which creates/downloads a zip file for a list of asset IDs. 

The topic has various aspects:
- It should be implemented as background process with the job execution engine. Take a look at this related issue for some details: https://github.com/pimcore/planning/issues/352. The actual implementation of the zip download job action can be done in the backend bundle as it will only be relevant for Pimcore studio.

- We need a progress bar while the zip generation is running.

- To inform the user about the progress (percentage) and completion of the zip file generation we should use Mercure to send status updates back to the user. Please talk to @alexz707 for details as he is already working on this topic.

- We should also think about the error handling and how to inform the user if something fails.

- What happens when the user closes his browser and later comes back and in the meantime the zip generation is finished?

As the topic involves multiple aspect it would be fine if we split it up into multiple sub-tickets. We will need a similar approach for multiple of these batch jobs, so we should think about the topic a little bit in a generic/holistic way. If you need some discussion about the topic @vin0401 and me will be happy to talk about it. :)

mattamon:
Resolves #137 

## Additional info
Adding GEE setup and asset download


config/pimcore/execution_engine.yaml
```diff
@@ -0,0 +1,12 @@
+pimcore_generic_execution_engine:
+  error_handling: 'continue_on_error'
```
lukmzig: I would say we should change this to `stop_on_first_error` :)

mattamon: Done :)

lukmzig (approved):
🦭

PR #156, opened 2024-06-19
Update schedule actions

mattamon:
Only use publish instead of publish-version to make it consistent with delete action


Corepex (approved):
LGTM

PR #160, opened 2024-06-20
[Task] Add metadata to image version

mattamon:
Part of https://github.com/pimcore/studio-ui-bundle/issues/352

### DIFF
--- a/src/Version/Schema/CustomMetadataVersion.php
+++ b/src/Version/Schema/CustomMetadataVersion.php
@@ -1,0 +1,60 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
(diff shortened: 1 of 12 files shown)


src/Version/Schema/CustomMetadataVersion.php
```diff
@@ -0,0 +1,60 @@
...
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Version\Schema;
+
+use OpenApi\Attributes\Property;
+use OpenApi\Attributes\Schema;
+
```
lukmzig, line 21: Can it be internal? :)

lukmzig (approved):
One tiny thing, the rest LGTM :)

src/Version/Schema/CustomMetadataVersion.php
mattamon: Good point since we do not give it to the outer world it can be internal yes :)

mattamon: Question is, should we remove it then from every schema that has the additional attributes, I guess so right?

lukmzig: Hmmm I guess that would make sense, yeah

PR #164, opened 2024-06-20
Make schemas public that use additional attributes

### LINKED ISSUE #163: [Task] Check schemas for addition attributes
Every schema that does implement the additional attributes should not be tagged as internal

mattamon:
Resolves #163 

## Additional info
Schemas that use the additional attributes need to be considered as public.

Using them in an event e.g. where you have an additional method with an input parameter of that schema, it will be shown as internal.


lukmzig (approved):
🦭

PR #169, opened 2024-06-24
[Task] Thumbnail for versions

### LINKED ISSUE #361: [Versions] Preview image thumbnails
The version preview images need to be thumbnails:

image

Currently it seems that the full image content get's displayed. This works for the small assets in the demo but will fail for big assets (or even strange formats like *.psd).

Related to #349

mattamon:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/361


src/Version/Service/VersionBinaryService.php
```diff
@@ -65,21 +67,29 @@ public function downloadAsset(
...
         $version = $this->repository->getVersionById($id);
-        $element = $this->repository->getElementFromVersion($version, $user);
-        if (!$element instanceof Asset\Image) {
-            throw new InvalidElementTypeException($element->getType());
+        $image = $this->repository->getElementFromVersion($version, $user);
+        if (!$image instanceof Asset\Image) {
+            throw new InvalidElementTypeException($image->getType());
+        }
+
+        $config = Config::getPreviewConfig();
```
lukmzig: Should we introduce a new resolver here? Or you think its not necessary?

mattamon: Ah yes, totally forgot about that for the review!
I will add a new resolver

lukmzig (commented):
In general it looks good to me :)
One question, should we also change the download endpoint to return only asset with the pre-configured thumbnail? I am not sure if we need to keep it consistent

mattamon:
> In general it looks good to me :) One question, should we also change the download endpoint to return only asset with the pre-configured thumbnail? I am not sure if we need to keep it consistent

Good question. I am also not sure about it. Maybe create an issue to keep that in mind.

lukmzig:
> > In general it looks good to me :) One question, should we also change the download endpoint to return only asset with the pre-configured thumbnail? I am not sure if we need to keep it consistent
> 
> Good question. I am also not sure about it. Maybe create an issue to keep that in mind.

New issue is here: https://github.com/pimcore/studio-backend-bundle/issues/170

lukmzig (approved):
🦭

PR #174, opened 2024-06-24
#162 [User Management] [Endpoint] Update Password

### LINKED ISSUE #162: [User Management] [Endpoint] Update Password
Part of https://github.com/pimcore/studio-ui-bundle/issues/180

martineiber:
Resolves #162


mattamon (approved):
LGTM 🦭

Maybe just as a thought, we should update our password requirements. 
We write the year 2024 https://www.upguard.com/blog/the-password-security-checklist and our only requirement is, that the password is at least 10 characters long.

PR #181, opened 2024-06-25
[Delete Info] Elements

### LINKED ISSUE #168: [Delete Info] Elements
Add a new endpoint to count children and return info if they will be added to recycle bin and check for dependency of the element

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/168

## Additional info
Add endpoint to return delete info about elements 
 - check if element has dependencies
 - check if recycle bin will be used (based on the configured threshold)


mattamon (approved):
🦭

PR #189, opened 2024-06-26
Restructure User and Role section

martineiber:
 - Move role into their own section 
 - Endpoint for the user tree view is now under /users/tree
 - Minor name changes


mattamon (approved):
🦭

PR #199, opened 2024-06-28
Add imports for standard php function for micro optimization

### LINKED ISSUE #200: Micro Optimization
Use imports for standard php functionality since it seems that they are slightly faster that way.

See https://www.deviaene.eu/articles/2023/why-prefix-php-functions-calls-with-backslash/

mattamon:
Resolves #200


mattamon:
Closed in favor of #201

PR #201, opened 2024-06-28
[Task] CS Fixer for Micro Optimiziation

mattamon:
See title

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -32,7 +32,6 @@ $config->setRules([
             ' @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)' . PHP_EOL .
             ' @license    http://www.pimcore.org/license     GPLv3 and PCL'
     ],
-
     'blank_line_before_statement'         => true,
     'encoding'                            => true,
     'function_typehint_space'             => true,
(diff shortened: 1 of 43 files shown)


.php-cs-fixer.dist.php
```diff
@@ -72,6 +72,7 @@
     'ternary_operator_spaces'             => true,
     'trailing_comma_in_multiline'         => true,
     'whitespace_after_comma_in_array'     => true,
+    'import_functions'                    => true,
```
jdreesen: See: https://github.com/pimcore/studio-backend-bundle/issues/200#issuecomment-2197184449

mattamon: Thanks for the comment! I will add it!

jdreesen: You can still use `'global_namespace_import' => ['import_functions' => true],` if you prefer imports to backslashes before every function call.

mattamon: yeah. I just readded it :) Thanks for the hint though. With all the new bundles we try to rather do use statements.

jdreesen: There's also [`ordered_imports`](https://mlocati.github.io/php-cs-fixer-configurator/#version:3.59|fixer:ordered_imports) if you want to separate/group class/const/function imports.

mattamon: even better, yeah i find it not so nice if they are mixed. thank you very much!

PR #203, opened 2024-07-01
#175 [User Management] [Endpoint] Get Roles/Folder for tree view.

### LINKED ISSUE #175: [User Management] [Endpoint] Get Roles/Folder for tree view.
Part of https://github.com/pimcore/studio-ui-bundle/issues/181

martineiber:
Resolves #175

Refactor TreeNode Schema, moved it to OpenApi Namespace.


src/Role/MappedParameter/RoleTreeListingParameter.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @internal
+ */
+final readonly class RoleTreeListingParameter
+{
+    public function __construct(
+        #[PositiveOrZero]
+        #[NotBlank]
+        private int $parentId = 1
```
mattamon: Do we need a whole class for Role here that just has the parentId in it? 
Not sure if we should make it more generic here.

WDYT?

mattamon (changes requested):
LGTM, maybe one small thing.

src/Role/MappedParameter/RoleTreeListingParameter.php
martineiber: I would say this is a good Idea, since we have this also in the User Section. Than we should make it something like ParentIdParameter. I will take a look.

mattamon (approved):
LGTM 🦭

PR #206, opened 2024-07-01
[Feature] Grid Functionality

### LINKED ISSUE #175: Asset List Grid Config

mattamon:
Partially resolves https://github.com/pimcore/studio-ui-bundle/issues/175


src/Grid/Column/ColumnDefinitionInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Grid\Column;
+
+/**
+ * @internal
+ */
```
martineiber: Should not be Internal?

martineiber:
```suggestion
```

src/Grid/Column/Resolver/DatetimeResolver.php
```diff
@@ -0,0 +1,61 @@
...
+ * @internal
+ */
+final class DatetimeResolver implements ColumnResolverInterface
+{
+    use SimpleGetterTrait;
+    use ColumnDataTrait;
+
+    /**
+     * @throws InvalidArgumentException
+     */
```
martineiber, line 38: Isn't Carbon::create throwing InvalidFormatException?

And what about handling the Exception here?  Like return null as data? WDYT?

src/Grid/Column/Resolver/FileSizeResolver.php
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * @internal
+ */
+final class FileSizeResolver implements ColumnResolverInterface
+{
+    use ColumnDataTrait;
+
+    /**
+     * @throws InvalidArgumentException
+     */
```
martineiber: Same here... Where is this exception coming from?

src/Grid/Util/Trait/SimpleGetterTrait.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Pimcore\Bundle\StudioBackendBundle\Grid\Util\Trait;
+
+use Pimcore\Bundle\StudioBackendBundle\Exception\Api\InvalidArgumentException;
+use Pimcore\Bundle\StudioBackendBundle\Grid\Schema\Column;
+use Pimcore\Model\Element\ElementInterface;
+
+/**
+ * @internal
+ */
```
martineiber: Could also be public?

martineiber (changes requested):
Some minor Questions :) Rest LGTM :+1:

src/Grid/Column/Resolver/DatetimeResolver.php
mattamon: Exception is coming from the SimpleGetter Trait, since if the field does not exist, it throws the invalidargument exception.
Same goes for all the other resolvers

mattamon: In general we should maybe create a follow up ticket and talk to the frontend on how to deliver dates

src/Grid/Column/Resolver/FileSizeResolver.php
martineiber: So this can be removed.. SInce it does not have the SimpleGetter?

src/Grid/Column/Resolver/DatetimeResolver.php
martineiber: But then we should add the InvalidFormatException as well here?

mattamon: Maybe we should just deliver the timestamp for now, not make things more complicated than they need to be, if we do not know how frontend will handle it anyways :)

martineiber: Works for me :) Sounds Good :+1:

martineiber (approved):
LGTM :+1:

PR #233, opened 2024-07-08
#220 [User Management] Update Role

### LINKED ISSUE #220: [User Management] Update Role
Part of https://github.com/pimcore/studio-ui-bundle/issues/181

martineiber:
Resolves #220 

## Additional info
Minor Refactoring to reuse Update Services.

### DIFF
--- a/src/Role/Controller/UpdateRoleController.php
+++ b/src/Role/Controller/UpdateRoleController.php
@@ -27,13 +27,11 @@ use Pimcore\Bundle\StudioBackendBundle\OpenApi\Attributes\Response\DefaultRespon
 use Pimcore\Bundle\StudioBackendBundle\OpenApi\Attributes\Response\SuccessResponse;
 use Pimcore\Bundle\StudioBackendBundle\OpenApi\Config\Tags;
 use Pimcore\Bundle\StudioBackendBundle\Role\MappedParameter\UpdateRoleParameter;
-use Pimcore\Bundle\StudioBackendBundle\Role\Schema\DetailedRole;
 use Pimcore\Bundle\StudioBackendBundle\Role\Schema\UpdateRole;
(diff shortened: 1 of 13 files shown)


src/Role/Controller/UpdateRoleController.php
```diff
@@ -0,0 +1,84 @@
...
+    #[RequestBody(
+        content: new JsonContent(ref: UpdateRole::class)
+    )]
+    #[DefaultResponses([
+        HttpResponseCodes::NOT_FOUND,
+    ])]
+    public function updateRoleById(int $id, #[MapRequestPayload] UpdateRoleParameter $roleUpdate): JsonResponse
+    {
+        return $this->jsonResponse(
+            $this->roleUpdateService->updateRoleById($id, $roleUpdate)
```
mattamon: For e.g. update the asset, we call the updateService, but it actually does not return anything.

The thing is, that we should not return the updated data immediately, because there could be save listeners to the object. It would be better to call the "get" again instead of returning the updated data directly. 
At least that was our way of thinking for the assets.
Should be consistent with the rest. 

This would also reduce code duplication in case of the additional attributes since it only then needs to be in the "get" operation

mattamon (changes requested):
Rest looks good to me :)

src/Role/Controller/UpdateRoleController.php
martineiber: Sounds reasonable! I will change it, and also for the Update User Endpoint.

martineiber: Fixed https://github.com/pimcore/studio-backend-bundle/pull/233/commits/591b2061b8c1d9734af536c53f27b62ce4a8b590

mattamon (approved):
LGTM

PR #235, opened 2024-07-08
[Task] Adding userid to finished event for SSE

### LINKED ISSUE #230: [Task] Adding userid to finished event for SSE
Finished messages should be user specific

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/230

### DIFF
--- a/src/Asset/EventSubscriber/DeletionSubscriber.php
+++ b/src/Asset/EventSubscriber/DeletionSubscriber.php
@@ -49,6 +49,7 @@ final readonly class DeletionSubscriber implements EventSubscriberInterface
         if ($event->getJobName() !==  Jobs::DELETE_ASSETS->value) {
             return;
         }
+        $ownerId = $event->getJobRunOwnerId();
 
         match ($event->getNewState()) {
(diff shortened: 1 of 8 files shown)


src/Asset/EventSubscriber/DeletionSubscriber.php
```diff
@@ -49,18 +49,21 @@ public function onStateChanged(JobRunStateChangedEvent $event): void
         if ($event->getJobName() !==  Jobs::DELETE_ASSETS->value) {
             return;
         }
+        $ownerId = $event->getJobRunOwnerId();
```
mattamon: I don't think an extra variable is necessary right? We have it all encapsulated in the event already.

WDYT?

mattamon (changes requested):
Rest gets the Seal

src/Asset/EventSubscriber/DeletionSubscriber.php
lukmzig: sounds good :)

PR #236, opened 2024-07-09
[Task] Introduce lifetime and hubservice

### LINKED ISSUE #185: [Task] Add cookie lifetime to mercure settings Follow up Mercure
Follow up to https://github.com/pimcore/studio-backend-bundle/issues/184

Add the possibility to add a lifetime for the JWT Token & cookie in the mercure settings

mattamon:
Resolves #185 

## Additional info
Refactoring cookie creation and introducing lifetime


src/Mercure/Service/HubService.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\Mercure\HubInterface;
+
+/**
+ * @internal
+ */
+final readonly class HubService implements HubServiceInterface
+{
+    public function __construct(
+        private HubInterface $clientHub,
+        private int $cookieLifetime = 3600
```
lukmzig, line 30: do we need this default value here, if the default value is set in the configuration?

mattamon: Yeah if it is not dependency injected it still has the same value, but yeah in general we could scrap it, but it also does not hurt imho

lukmzig (approved):
LGTM :seal:

Issue #244, opened 2024-07-10
[Task]: Verify we are using Flysystem everywhere


alexz707:
### Studio API version

1

### Steps to reproduce

Currently there are some implementations like unlink, fopen, ... which are not compatible with Flysystem (https://flysystem.thephpleague.com/docs/)

### Actual Behavior

We need to make sure to only use the Flysystem access methods so there is no problem if someone wants to use a different filesystem than the local one.

### Expected Behavior

to be fully compatible with Flysystem

dpfaffenbauer:
*Flysystem with a lower case `s` ;)

mattamon:
@dpfaffenbauer there you go :P changed it

PR #252, opened 2024-07-11
Refactor Download and Upload to respect flysystem

### LINKED ISSUE #245: [Task]: Downloads should be only accessible by the user which started the download
### Studio API version

1

### Steps to reproduce

Start a download process, get the link and try to retrieve it with another logged in pimcore user.

### Actual Behavior

The download starts, so the user gets files which he maybe is not allowed to see on his own.

### Expected Behavior

The download will not start because the user is not the one who started the download.

mattamon:
Resolves 
- #245 
- #244 
- https://github.com/pimcore/studio-backend-bundle/issues/256
- https://github.com/pimcore/studio-backend-bundle/issues/255

## Additional info
- Switching to jobrun id to get the owner to make permission check
- Make zip export processing sequential
- Make csv export processing sequential
- Use flysystem storage for resources


src/Asset/ExecutionEngine/AutomationAction/Messenger/Handler/CsvCreationHandler.php
```diff
@@ -59,87 +52,47 @@ public function __construct(
...
-
-        $csv = $this->csvService->getCsvFile($jobRun->getId(), $configuration, $settings);
-
-        if (!$csv) {
+        $assetData = $jobRun->getContext()[Csv::ASSET_EXPORT_DATA->value];
+
+        try {
+            $this->csvService->createCsvFile(
+                $jobRun->getId(),
+                $settings[Csv::SETTINGS_DELIMITER->value] ?? ',',
```
martineiber: Should we somewhere configure a default delimiter?

src/Asset/ExecutionEngine/AutomationAction/Messenger/Handler/CsvDataCollectionHandler.php
```diff
@@ -0,0 +1,132 @@
...
+use Pimcore\Bundle\StudioBackendBundle\Util\Traits\HandlerProgressTrait;
+use Pimcore\Model\Asset;
+use Pimcore\Model\Element\ElementDescriptor;
+use Symfony\Component\Messenger\Attribute\AsMessageHandler;
+
+/**
+ * @internal
+ */
+#[AsMessageHandler]
+final class CsvDataCollectionHandler extends AbstractHandler
```
martineiber, line 38: This can not be readonly?

src/DependencyInjection/Configuration.php
```diff
@@ -167,4 +169,22 @@ private function addMercureConfiguration(ArrayNodeDefinition $node): void
                 ->end()
             ->end();
     }
+
+    private function addAssetDownloadLimits(ArrayNodeDefinition $node): void
+    {
+        $node->children()
```
martineiber, line 176: Can you maybe add a short documentation about this configuration, with an example?

martineiber (changes requested):
Just some small remarks. Rest :seal:

src/Asset/ExecutionEngine/AutomationAction/Messenger/Handler/CsvDataCollectionHandler.php
lukmzig: No, because it extends a non - readonly classes AbstractHandler and AbstractAutomationActionHandler

src/Asset/ExecutionEngine/AutomationAction/Messenger/Handler/CsvCreationHandler.php
lukmzig: Not sure, this should come from the grid config afaik ... what would you suggest?

martineiber: It's about if there is nothing from the config. So the ?? ',' path. 
If we should move the ',' somewhere in a config.

martineiber (approved):
LGTM :seal:

PR #259, opened 2024-07-15
#253 [Grid] Differentiate between ColumnConfiguration and ColumnData

### LINKED ISSUE #253: [Grid] Get Grid Data should not use same Schema
Part of https://github.com/pimcore/studio-ui-bundle/issues/175

Get Data should not use same Schema as Grid Configuration 

Example Endpoint: 
`studio/api/assets/grid`

Should be a minimised Version 
sortable, editable is not necessary.

martineiber:
Resolves #253

## Additional info

On the request level we now have three main components for the grid: `Column`, `ColumnConfiguration` and `ColumnData`.

### DIFF
--- a/doc/03_Grid.md
+++ b/doc/03_Grid.md
@@ -3,11 +3,11 @@
 On the request level we have three main components for the grid: `Column`, `ColumnConfiguration` and `ColumnData`.
 
 ## Column
-A column is a single column in the grid. It has a name, type adn a local. It is used to get the data for the column.
+A column is a single column in the grid. It has a name, type and a locale. It is used to get the data for the column.
 in addition, it has a configuration which can be used to configure the column, like the direction of the sorting
 
 ## ColumnConfiguration
-A column configuration is how the column should behave, for example if it should be sortable or editable. 
+A column configuration represents how the column should behave, for example if it should be sort-able or editable. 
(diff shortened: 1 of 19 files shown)


doc/03_Grid.md
```diff
@@ -0,0 +1,13 @@
+# Grid
+
+On the request level we have three main components for the grid: `Column`, `ColumnConfiguration` and `ColumnData`.
+
+## Column
+A column is a single column in the grid. It has a name, type adn a local. It is used to get the data for the column.
```
lukmzig:
```suggestion
A column is a single column in the grid. It has a name, type and a locale. It is used to get the data for the column.
```

doc/03_Grid.md
```diff
@@ -0,0 +1,13 @@
...
+in addition, it has a configuration which can be used to configure the column, like the direction of the sorting
+
+## ColumnConfiguration
+A column configuration is how the column should behave, for example if it should be sortable or editable.
```
lukmzig:
```suggestion
A column configuration represents how the column should behave, for example if it should be sort-able or editable. 
```

lukmzig (approved):
Some tiny changes in docs maybe
The rest :seal:

PR #268, opened 2024-07-22
Added path prefix `assign` to routes in `AssignController`

Corepex:
To fix the collision of route paths I added the a prefix (`assign`) to the path.

image

### DIFF
--- a/src/Tag/Controller/Element/AssignController.php
+++ b/src/Tag/Controller/Element/AssignController.php
@@ -48,10 +48,10 @@ final class AssignController extends AbstractApiController
     /**
      * @throws ElementSavingFailedException|NotFoundException
      */
-    #[Route('/tags/{elementType}/{id}/{tagId}', name: 'pimcore_studio_api_assign_element_tag', methods: ['POST'])]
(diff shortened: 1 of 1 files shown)


src/Tag/Controller/Element/AssignController.php
```diff
@@ -48,10 +48,10 @@ public function __construct(
     /**
      * @throws ElementSavingFailedException|NotFoundException
      */
-    #[Route('/tags/{elementType}/{id}/{tagId}', name: 'pimcore_studio_api_assign_element_tag', methods: ['POST'])]
+    #[Route('/tags/assign/{elementType}/{id}/{tagId}', name: 'pimcore_studio_api_assign_element_tag', methods: ['POST'])]
```
martineiber, line 51: Sonar is now complaining that the line is too long :D

Pls split it up in multiple lines..

martineiber: Sonar is now complaining that the line is too long :D

Pls split it up in multiple lines..

martineiber (changes requested):

src/Tag/Controller/Element/AssignController.php
Corepex: done :)

martineiber (approved):
LGTM :+1:

PR #270, opened 2024-07-22
Refactor Filter structure.

martineiber:
Resolves #

## Additional info
Refactor Filter structure that the filters are reusable in other request as well and not only in the Asset Collection. Currently the Filter are depending on the CollectionParametersInterface. 

Now each Filter has its own Interface and those can be combined to build a ParameterClass like here
[ElementParameters.php](https://github.com/pimcore/studio-backend-bundle/compare/refactor-filter-structure?expand=1#diff-46e4d6fcc3f64a51d0e19c2e9fc66ef5f511a5256f6f6489ef7df4dc44f9c40b)


lukmzig (approved):
:seal:

PR #271, opened 2024-07-22
Version preview exception

### LINKED ISSUE #267: Version preview exception
When I create a version e.g. for Asset ID 254 of the demo-enterprise and then call the versions preview endpoint the following exception appears:

image

I guess this happens because of the asset metadata:

image

There we have linked cars - can they be the reason that the serialization fails?

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/267


martineiber (approved):
LGTM

PR #274, opened 2024-07-24
[Task]: Downloads should be able to be aborted & deleted

### LINKED ISSUE #246: [Task]: Downloads should be able to be aborted & deleted
1. The download process should have the possibility to be aborted.
So steps which are maybe generated should also check if the still need to run or if the can be stopped.

2. There should be an endpoint which can delete a download (without relying on the cleanup process)

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/246

## Additional info
- add endpoint to abort the jobRunAction
- add endpoint to cleanup download files


src/ExecutionEngine/Controller/AbortController.php
```diff
@@ -0,0 +1,74 @@
...
+        tags: [Tags::ExecutionEngine->name]
+    )]
+    #[IdParameter(type: 'JobRun', name: 'jobRunId')]
+    #[SuccessResponse]
+    #[DefaultResponses([
+        HttpResponseCodes::FORBIDDEN,
+        HttpResponseCodes::NOT_FOUND,
+        HttpResponseCodes::UNAUTHORIZED,
+    ])]
+    public function createZippedAssets(
```
martineiber:
```suggestion
    public function abortJobRun(
```

src/Asset/Controller/Download/DeleteCsvController.php
```diff
@@ -0,0 +1,78 @@
...
+        tags: [Tags::Assets->name]
+    )]
+    #[IdParameter(type: 'JobRun', name: 'jobRunId')]
+    #[SuccessResponse]
+    #[DefaultResponses([
+        HttpResponseCodes::UNAUTHORIZED,
+        HttpResponseCodes::FORBIDDEN,
+        HttpResponseCodes::NOT_FOUND,
+    ])]
+    public function deleteAssetsZip(int $jobRunId): Response
```
martineiber:
```suggestion
    public function deleteAssetsCsv(int $jobRunId): Response
```

martineiber (changes requested):

martineiber (approved):
LGTM :seal:

PR #275, opened 2024-07-24
[Notes&Event] User should be saved when creating a new note

### LINKED ISSUE #272: [Notes&Event] User should be saved when creating a new note
When adding a note via the notes API the user of the note should be set to the authenticated user. Currently the userId and userName are null when a new note is added and then the note list is fetched again.

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/272


martineiber (approved):
LGTM :seal:

Issue #261, opened 2024-07-16
[Grid] Add Filter & Search


martineiber:
Part of https://github.com/pimcore/studio-ui-bundle/issues/176

PR #277, opened 2024-07-25
#261 [Grid] Add MetaData Filter

martineiber:
Part of #261


src/DataIndex/Query/AssetQuery.php
```diff
@@ -97,4 +98,11 @@ public function searchByIds(array $ids): self
 
         return $this;
     }
+
+    public function filterMetaDate(string $name, string $type, mixed $data): self
```
lukmzig:
```suggestion
    public function filterMetaData(string $name, string $type, mixed $data): self
```

src/DataIndex/Filter/Asset/MetaData/TextAreaFilter.php
```diff
@@ -0,0 +1,60 @@
...
+        return $query;
+    }
+
+    private function applyTextAreaFilter(ColumnFilter $column, AssetQuery $query): AssetQuery
+    {
+        if (!is_string($column->getFilterValue())) {
+            throw new InvalidArgumentException('Filter value for textarea must be a string');
+        }
+
+        $query->filterMetaDate($column->getKey(), FilterType::TEXTAREA->value, $column->getFilterValue());
```
lukmzig:
```suggestion
        $query->filterMetaData($column->getKey(), FilterType::TEXTAREA->value, $column->getFilterValue());
```

src/DataIndex/Query/AssetQuery.php
martineiber: Good Catch 

Fixed here [a376936](https://github.com/pimcore/studio-backend-bundle/pull/277/commits/a3769367344122ff8ec284fbfd6b5a227ee3cb17)

lukmzig (approved):
image

PR #280, opened 2024-07-26
[Bug] Fix Type check.

martineiber:
Wrong type check... Return null if type does NOT match


lukmzig (approved):
:seal:

PR #284, opened 2024-07-29
#261 [Grid] Add System field Filter

### LINKED ISSUE #261: [Grid] Add Filter & Search

martineiber:
Resolves #261 

## Additional info

https://github.com/pimcore/generic-data-index-bundle/pull/185
https://github.com/pimcore/generic-data-index-bundle/pull/194
Needs to be merged first.


src/DataIndex/Filter/Asset/MetaData/IsAssetFilterTrait.php
```diff
@@ -22,7 +22,7 @@
 /**
  * @internal
  */
-trait IsAssetMetaDataTrait
+trait IsAssetFilterTrait
```
lukmzig: Not sure if its just displayed like this in github, but should we move this trait out of the MetaData namespace?

lukmzig (changes requested):
One small thing, the rest LGTM :seal:

src/DataIndex/Filter/Asset/MetaData/IsAssetFilterTrait.php
martineiber: Good Point.. I will move it :)

lukmzig (approved):

PR #293, opened 2024-07-31
[Custom Metadata] Serializer exception

### LINKED ISSUE #290: [Custom Metadata] Serializer exception
We had a similar case also in the versions API: #267

When a asset metadata attribute contains multiple cars we get an exception like this:

image

Not sure if this needs to be solved in the backend bundle or will be part of the asset metadata class definitions implementation? The multi releation data type is part of this bundle but not of the core.

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/290


martineiber (approved):
:seal:

PR #308, opened 2024-08-06
[Improvement] Add exportable flag to column definition 

### LINKED ISSUE #258: [Grid] Definition should configure if it is exportable or not
Add a "isExportable" flag to the definition to be used for the CSV Export.

mattamon:
Resolves #258 

## Additional info
Not every column is exportable into e.g csv

### DIFF
--- a/src/Grid/Service/GridService.php
+++ b/src/Grid/Service/GridService.php
@@ -136,6 +136,7 @@ final class GridService implements GridServiceInterface
         string $elementType
     ): array {
         $data = $this->getGridDataForElement($columnCollection, $element, $elementType);
+
         return array_map(
             static fn (ColumnData $columnData) => $columnData->getValue(),
             $data['columns']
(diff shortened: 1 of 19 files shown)


src/Grid/Service/GridService.php
```diff
@@ -268,4 +272,13 @@ private function getColumnResolvers(): array
 
         return $this->columnResolvers;
     }
+
+    private function isExportable(string $type): bool
+    {
+        if (!array_key_exists($type, $this->getColumnDefinitions())) {
+            return false;
```
mattamon, line 275: Should we maybe already throw an exception here?

martineiber: I would not throw it here i would do it instead of the `continue`. 

It would be a better feedback for the user that he knows why some columns are not in the exported file... So i would work with an exception here.

mattamon: Sounds good to me!

martineiber (approved):
:seal:

PR #309, opened 2024-08-06
[Icons] Data object tree - add icons

### LINKED ISSUE #304: [Icons] Data object tree - add icons
We should add icons to the tree endpoint
There are in general 2 types
- default icons
- icons from the class definition

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/304


mattamon (approved):
Due to the extreme heat, the seal has the week off. 

The 🐢 is taking over for the week. 

Why did the tortoise start giving out seals of approval?

Because slow and steady wins the race, but a tortoise's approval is really worth the wait! 🐢

PR #310, opened 2024-08-06
[Grid] Add localization

### LINKED ISSUE #299: [Grid] localization for grid data.
take into account the localization for the grid data.

mattamon:
Resolves #299 

## Additional info
Add localization to metadata and system.string


src/Grid/Util/Trait/MetaData/LocalizedValueTrait.php
```diff
@@ -0,0 +1,37 @@
...
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Grid\Util\Trait\MetaData;
```
martineiber, line 17: Since this is only for Assets. We should move it into a Asset Namespace? Or Call it AssetMetaData?
And add a Check if $element is an instance of Asset?

martineiber (changes requested):

src/Grid/Util/Trait/MetaData/LocalizedValueTrait.php
mattamon: I would not move it to the asset namespace since it is clearly only a grid functionality, with the column parameter. 
I already moved it to the subnamespace metadata but of course we could  rename it to AssetMetaData. But to be consistent also in the Resolver we should then rename it, cause there it is just called MetaData. I will add the check, good catch.

src/Grid/Util/Trait/MetaData/LocalizedValueTrait.php
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * @internal
+ */
+trait LocalizedValueTrait
+{
+    private function getLocalizedValue(Column $column, ElementInterface $element): ?string
+    {
+        if (!$element instanceof Asset) {
+            return null;
```
martineiber: Just an Idea... Should we not throw an Exception here?

mattamon: Hmm, so you would need to call e.g. the object grid and put metadata.input or metadata.textarea as type. Yeah that is kinda a hard missconfiguration. We can throw an exception here too then. InvalidArgument should do it? WDYT?

mattamon: Done in https://github.com/pimcore/studio-backend-bundle/pull/310/commits/0d4ea9cb0f83c27af9ccd4546eb2a297a7b29c09

martineiber (approved):
:seal:

PR #311, opened 2024-08-06
#297 [Grid] Default grid configuration

### LINKED ISSUE #297: [Grid] Default configuration
We need an Endpoint which delivers the default grid configuration for assets.

martineiber:
Resolves #297 

## Additional info
Contains some refactoring to ConfigurrationService from the Grid Service. And some Controller renaming.


src/Asset/Controller/Grid/Configuration/GetConfigurationController.php
```diff
@@ -0,0 +1,98 @@
...
+        summary: 'Get asset grid configuration for a specific folder',
+        tags: [Tags::Grid->name]
+    )]
+    #[IdParameter(name: 'folderId')]
+    #[IdParameter(name: 'configurationId', required: false)]
+    #[SuccessResponse(
+        description: 'Grid configuration',
+        content: new JsonContent(
+            properties: [
+                new Property(
```
mattamon, line 74: Don't we have a generic version of an array property somewhere? I am not sure though cause we already have so many files :D

src/Grid/Service/ConfigurationService.php
```diff
@@ -0,0 +1,94 @@
...
+    public function getAvailableAssetGridConfiguration(): array
+    {
+        $columns = [];
+        foreach ($this->gridService->getColumnCollectors() as $collector) {
+            // Only collect supported asset collectors
+            if (!in_array(ElementTypes::TYPE_ASSET, $collector->supportedElementTypes(), true)) {
+                continue;
+            }
+
+            $columns = array_merge(
```
mattamon: this can be replaced by the spread operator like here https://github.com/pimcore/studio-backend-bundle/pull/308/files#diff-614133628b0e63d8f255992a27087738b4073af6b3e1cccd5ce6d886bbace99dR159

Array merge is very greedy, even PHPstorm complains :D

src/Grid/Service/ConfigurationService.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public function getDefaultAssetGridConfiguration(): array
+    {
+        $availableColumns = $this->getAvailableAssetGridConfiguration();
+        $defaultColumns = [];
+        foreach ($this->predefinedColumns as $predefinedColumn) {
+            $filteredColumns =
+                array_filter($availableColumns, function (ColumnConfiguration $column) use ($predefinedColumn) {
+                    if ($column->getKey() === $predefinedColumn['key'] &&
+                        $column->getGroup() === $predefinedColumn['group']
```
mattamon, line 78: can't this be just replaced by `return $column->getKey() === $predefinedColumn['key'] &&
                        $column->getGroup() === $predefinedColumn['group']` ?

mattamon (commented):
Some small questions, otherwise looks good!

src/Grid/Service/ConfigurationService.php
mattamon: I guess with the merge of the export, you have to check this anyways again. Sorry :)

src/Asset/Controller/Grid/Configuration/GetConfigurationController.php
martineiber: For this case i could not found any generic version of it.

src/Grid/Service/ConfigurationService.php
martineiber: Yeah good point... I just moved it from the GridService... but i will change it.

src/Grid/Service/ConfigurationService.php
martineiber: Yes could be replaced... but IMHO it is better to read as it is now!

WDYT?

mattamon: Puh, I really don't mind both.
Logically of course if you already evaluate a condition and could use this a the return value already I would not introduce an extra if, if the conditions gets evaluated anyways but we can keep it as is of course.

src/Grid/Service/ConfigurationService.php
martineiber: Fixed https://github.com/pimcore/studio-backend-bundle/pull/311/commits/bd0ded320676342dc5acda928cdbb827c3a489ac

mattamon (approved):
🦭

PR #313, opened 2024-08-07
[Task] Make spelling more consistent with metadata

### LINKED ISSUE #312: [Task] Make Metadata spelling consistent
metadata should not be cased since it is its own word.

mattamon:
Resolves #312 

## Additional info
Cannot rename everywhere, since it used for schema in the api


martineiber (approved):
:seal:

src/Version/Schema/ImageVersion.php
```diff
@@ -42,7 +42,7 @@ public function __construct(
         private readonly int $fileSize,
         #[Property(description: 'mime type', type: 'string', example: 'image/png')]
         private readonly string $mimeType,
-        #[Property(description: 'Metadata', type: 'array', items: new Items(ref: CustomMetadataVersion::class))]
+        #[Property(description: 'Metadata', type: 'array', items: new Items(ref: CustomMetaDataVersion::class))]
```
lukmzig: Should this be also 'MetaData'?

src/Asset/Attributes/Property/UpdateCustomMetaData.php
```diff
@@ -28,9 +28,9 @@ final class UpdateCustomMetaData extends Property
     public function __construct()
     {
         parent::__construct(
-            'metadata',
+            property: 'metadata',
```
lukmzig, line 31: should we also adapt the property names?

src/Asset/Attributes/Property/CustomMetaData.php
```diff
@@ -18,17 +18,17 @@
...
  * @internal
  */
 final class CustomMetaData extends Property
 {
-    public function __construct(string $schema = UpdateCustomMetadataSchema::class)
+    public function __construct(string $schema = UpdateCustomMetaDataSchema::class)
     {
         parent::__construct(
-            'metadata',
+            property: 'metadata',
```
lukmzig, line 31: should we also adapt the property names?

lukmzig (commented):
Some small things, the rest :seal:

src/Version/Schema/ImageVersion.php
mattamon: Yeah we should also change the description good catch!

src/Asset/Attributes/Property/UpdateCustomMetaData.php
mattamon: I would not, since this is used by the serializer and I would not change the API property names?
Or should we?

lukmzig: Not sure to be honest. For custom settings fro example we use customSettings ...so maybe just to keep it consistent?

src/Asset/Service/Data/CustomMetaDataServiceInterface.php
```diff
@@ -16,18 +16,19 @@
...
  * @internal
  */
-interface CustomMetadataServiceInterface
+interface CustomMetaDataServiceInterface
 {
     /**
-     * @throws AccessDeniedException
+     * @return array<int, CustomMetaData>
+     *
+     *@throws AccessDeniedException
```
martineiber:
```suggestion
     * @throws AccessDeniedException
```

martineiber (changes requested):

src/Asset/Attributes/Property/UpdateCustomMetaData.php
mattamon: I would even prefer it the other way round to write everything small metadata, but that are also more changes, because it is actually one word.

mattamon: @lukmzig @martineiber should I turn it around and make it metadata everywhere instead of metaData since it is actually its own word?

lukmzig: maybe that would make more sense?
since I guess in the core its also called metadata based on this:
https://github.com/pimcore/studio-backend-bundle/pull/313/files#diff-69c196c7cd38d69b327debc1f015c2d583aa9f9f20c753ca2839b74ed2b4aad7R34

martineiber:
DJ mattamon in the house 🎤

src/Asset/Attributes/Property/UpdateCustomMetaData.php
mattamon: Resolved https://github.com/pimcore/studio-backend-bundle/pull/313/commits/cc45a2b9a08918fc423ada5cffd9cc2ffa89ecc0

src/Asset/Service/Data/CustomMetadataService.php
```diff
@@ -44,9 +44,10 @@ public function __construct(
     }
 
     /**
-     * @throws AccessDeniedException
-     *
      * @return array<int, CustomMetadata>
+     *
+     *@throws AccessDeniedException
```
martineiber:
```suggestion
     * @throws AccessDeniedException
```

martineiber (changes requested):

src/Asset/Service/Data/CustomMetadataService.php
martineiber: Somehow this is again here :D

martineiber (approved):

lukmzig (approved):

PR #314, opened 2024-08-07
[Grid] Configuration ID as Query Parameter.

martineiber:
The FE Client can not work with optional path parameter. So this PR changes the Configuration ID to an Query Parameter.


mattamon (approved):
🦭

lukmzig (approved):
🦭

PR #315, opened 2024-08-07
Fix Localized Return type

martineiber:
Return type should be mixed for getLocalizedValue


mattamon (approved):
Spanish: aprobado
French: approuvé
German: genehmigt
Italian: approvato
Portuguese: aprovado

Approved in 6 different languages!

PR #316, opened 2024-08-07
Data Object Tree - Copy & Paste

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/264

## Additional info
Add endpoint to clone data objects
- clone single object
- clone objects recursively with GEE
- clone objects and update references (separate GEE job)


src/DataObject/Controller/CloneController.php
```diff
@@ -0,0 +1,105 @@
...
+        HttpResponseCodes::UNAUTHORIZED,
+        HttpResponseCodes::NOT_FOUND,
+    ])]
+    public function cloneDataObject(
+        int $id,
+        int $parentId,
+        #[MapRequestPayload] CloneParameters $parameters
+    ): Response {
+        $status = HttpResponseCodes::SUCCESS->value;
+        $data = null;
```
mattamon: Does data have a purpose here?
It is always going to be null no?

src/DataObject/Service/ExecutionEngine/CloneService.php
```diff
@@ -0,0 +1,179 @@
...
+                sprintf(
+                    'Missing permissions on target element %s',
+                    $parent->getId()
+                )
+            );
+        }
+
+        $this->synchronousProcessingService->enable();
+
+        $dataObject = (new DataObjectService())->copyAsChild(
```
mattamon: Can we use the resolver here?

mattamon (changes requested):
That looks like a lot of work went into this. : O

src/DataObject/Service/ExecutionEngine/CloneService.php
lukmzig: Not sure how since this is not a registered service

src/DataObject/Controller/CloneController.php
lukmzig: Good point, we can clean this up :)

src/DataObject/Service/ExecutionEngine/CloneService.php
mattamon: Core is somtimes mind boggling :D

mattamon (approved):
A utomatic 
P rocessing 
P rocedure 
R eview 
O peration 
V erification 
E valuation 
D one

PR #321, opened 2024-08-07
[Asset Patch] Add possibility to add predefined metadata in patch

### LINKED ISSUE #307: [Asset Patch Update] Metadata is not updated after a successful request
**Description**

Sending an patch request will result in a 200 Response, but after requesting the asset again, the asset data is not changed.

**Reproduction**

Example Payload:

```
{"data":[{"id":296,"metadata":[{"name":"author","data":"Peter Griffin"}]}]}
```

mattamon:
Resolves #307 

## Additional info
Before it was not possible to add metadata in patch


lukmzig (approved):
:eyes: good to me! :seal:

PR #325, opened 2024-08-08
[Task] Add language matching

### LINKED ISSUE #323: [Patch Metadata] Make patching multiple language entries possible
Currently only the name is checked, but name can be the same. Also the language needs to be checked.

mattamon:
Resolves #323 

## Additional info
Extend the matching also to the language instead of only the name


lukmzig (approved):
:seal:

Issue #326, opened 2024-08-08
[Api Translate][Task] Change summaries and descriptions to translation keys


mattamon:
As mentioned in #303 we can move the descriptions in the translations to make it more structured and removed text from the controller.

We should do this bit by bit. 

use the operationid from the method and convert it to snake case.

e.g. "cloneAsset" description then looks like clone_asset_description

Every description for summary property in the api is translateable
```[tasklist]
### Grouped by namespace
- [x] Asset
- [x] Authorization
- [x] DataObject
- [x] Element
- [x] Email
- [x] ExecutionEngine
- [x] Mercure
- [x] Note
- [x] Property
- [x] Role
- [x] Schedule
- [x] Settings
- [x] Tag
- [x] Thumbnail
- [x] Translation
- [x] User
- [x] Version
- [x] Workflow
```

PR #335, opened 2024-08-12
[Api Translate] Update operation ids in controllers

lukmzig:
Part of https://github.com/pimcore/studio-backend-bundle/issues/326


mattamon (approved):
🐢

PR #357, opened 2024-08-14
[Improvement] Introduce listing filter 

### LINKED ISSUE #351: [Proposal] Unify filtering for Listings and database queries
Currently for OpenSearch we use a tagged filter service approach

We should also use this approach for listings and database queries.

In the core, many filters were based on how extjs worked. This results in actually almost the same code written a few hundred times, just to have filtering. We should not do this anymore.

A good approach is to do it like for the grid. Chaining filters and adding these filters to either the listing or database e.g.

mattamon:
Resolves #351 

## Additional info
Mapper still needs to be improved


lukmzig (approved):
:otter:  -ly cool filtering

martineiber (approved):
🦦

PR #358, opened 2024-08-19
Notifications - Get One Notification

### LINKED ISSUE #353: Notifications - Get One Notification
Part of https://github.com/pimcore/studio-ui-bundle/issues/252
Add endpoint to get one notification data

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/353

Needs https://github.com/pimcore/pimcore/pull/17485


src/Notification/Service/NotificationService.php
```diff
@@ -54,8 +89,8 @@ public function listNotifications(FilterParameter $parameters): Collection
         foreach ($listing as $listEntry) {
             $entry = $this->notificationHydrator->hydrate($listEntry);
             $this->eventDispatcher->dispatch(
-                new NotificationEvent($entry),
-                NotificationEvent::EVENT_NAME
+                new NotificationListEvent($entry),
```
mattamon, line 92: Why do we need two different events here?

lukmzig: Because they work with different elements (NotificationListItem and Notification). NotificationListItem contains only data for the tree

src/Notification/Hydrator/NotificationHydrator.php
```diff
@@ -39,6 +40,23 @@ public function hydrate(Notification $notification): NotificationListItem
...
 
+    public function hydrateDetail(NotificationModel $notification): Notification
+    {
+        return new Notification(
+            $notification->getId(),
+            $notification->getType(),
+            $notification->getTitle(),
+            $notification->isRead(),
+            (bool)$notification->getLinkedElementType(),
+            (new Carbon($notification->getCreationDate(), 'UTC'))->getTimeStamp(),
```
martineiber, line 51: Should we add Carbon to the composer requirements? Since it is used by the Bundle, to be consistent. 
WDYT?

martineiber: Not sure if we have here any Rules here. Or if it is okay when it is required by the core.

mattamon: Yeah, good question. I mean it does not work without the core, but the bundle itself, should also list its requirements. So I would add it tbh.

lukmzig: I will add it then :)

martineiber (approved):
:penguin: loves notifications

PR #374, opened 2024-08-20
Notifications - Send Notification

lukmzig:
Part of  https://github.com/pimcore/studio-ui-bundle/issues/251


src/Notification/Controller/ReadController.php
```diff
@@ -48,7 +48,12 @@ public function __construct(
     /**
      * @throws AccessDeniedException|NotFoundException|UserNotFoundException
      */
-    #[Route('/notifications/{id}', name: 'pimcore_studio_api_read_notification', methods: ['POST'])]
+    #[Route(
+        '/notifications/{id}',
+        name: 'pimcore_studio_api_read_notification',
+        requirements: ['id' => '\d+'],
```
mattamon, line 54: This is new no? Should we introduce that for every path id?

mattamon (approved):
🦦 👀 see
🦦 👀 ✅

src/Notification/Controller/ReadController.php
lukmzig: We already have it e.g here: https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Asset/Controller/GetController.php#L52

it just ensures that route is not overwriting sub - routes (like here, read route would overwrite the send). I would say we should use it only when necessary

PR #379, opened 2024-08-21
[CustomSettings]: Renamed meta data field

### LINKED ISSUE #377: [Assets] Difference fixedCustomSettings and dynamicCustomSettings?
* What's the difference between `fixedCustomSettings` and `dynamicCustomSettings`?
* Why do I get the `embeddedMetadata` and `embeddedMetaDataExtracted` in both nodes with different values?

image

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/377

issue introduced by https://github.com/pimcore/studio-backend-bundle/pull/313
in the core, the custom settings inside of meta data are using MetaData


mattamon (approved):
🦦

PR #383, opened 2024-08-23
#367 [Grid] Update configuration

### LINKED ISSUE #367: [Grid] Update grid configuration

martineiber:
Resolves #367


src/Asset/MappedParameter/Grid/ColumnsAsArrayTrait.php
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @internal
+ */
+trait ColumnsAsArrayTrait
+{
+    public function getColumnsAsArray(): array
+    {
+        return array_map(
+            fn (ColumnSchema $column) => $column->toArray(),
```
mattamon:
```suggestion
            static fn (ColumnSchema $column) => $column->toArray(),
```

src/Entity/Grid/GridConfiguration.php
```diff
@@ -68,10 +68,20 @@ class GridConfiguration
     #[ORM\Column(type: 'datetime')]
     private DateTime $modificationDate;
 
-    #[ORM\OneToMany(mappedBy: 'configuration', targetEntity: GridConfigurationShare::class, cascade: ['persist'])]
+    #[ORM\OneToMany(
```
mattamon, line 71: Btw do we need that ORM\ or can we make a full import here too?

mattamon (changes requested):
🐧 says that the rest looks good!

src/Entity/Grid/GridConfiguration.php
martineiber: No we do not need the ORM\\.
We import the ORM like this `use Doctrine\ORM\Mapping as ORM;` IMHO i like it better that we know it has something todo with the ORM directly when we use it. Otherwise we have to import each ORM Mapping it self.

src/Asset/MappedParameter/Grid/ColumnsAsArrayTrait.php
martineiber: Good catch thx :)

src/Entity/Grid/GridConfiguration.php
mattamon: Sounds good!

mattamon (approved):
🐧 waddles for approval

lukmzig (approved):
🐧

PR #389, opened 2024-08-27
[Task] Send notifications for events which need a followup action

### LINKED ISSUE #247: [Task] Extend Notifications from Core
When a user starts a long running action and closed the browser the process continues in the background.
On the next day when the user logs into pimcore from a different browser he should get the needed notifications which are important to him. In this case that the download is ready.

Mercure has the feature to replay the subscribed topics from a given id.
This has two problems, the first one is that the user does not need all events from that time until now. (e.g he doesn't care about the progress of the download if it is already ready)
The second problem is that if there are many events they get overwritten so chances are high that the user misses some events.

Maybe a DB table is the way to go to only save "events" which are relevant to the user. So which need a follow up action.
This needs to be discussed.

lukmzig:
Resolves #247 

## Additional info
Send notifications with payload information after jobRun is finished and action is required by user (e.g downloads)


mattamon (approved):
🐧 🥇

PR #390, opened 2024-08-27
#382 [Grid] Get Configuration

### LINKED ISSUE #382: [Grid] Get specific configuration by ID

martineiber:
Resolves #382

### DIFF
--- a/src/Asset/Controller/Grid/Configuration/GetAvailableColumnsController.php
+++ b/src/Asset/Controller/Grid/Configuration/GetAvailableColumnsController.php
@@ -38,7 +38,7 @@ use Symfony\Component\Serializer\SerializerInterface;
 /**
  * @internal
  */
-final class GetAvailableConfigurationController extends AbstractApiController
+final class GetAvailableColumnsController extends AbstractApiController
 {
     public function __construct(
         SerializerInterface $serializer,
@@ -51,13 +51,13 @@ final class GetAvailableConfigurationController extends AbstractApiController
      * @throws NotFoundException|SearchException
      */
     #[Route(
-        '/assets/grid/available-configuration',
+        '/assets/grid/available-columns',
         name: 'pimcore_studio_api_get_asset_grid_available_configuration',
         methods: ['GET']
     )]
     #[IsGranted(UserPermissions::ASSETS->value)]
     #[Get(
-        path: self::API_PATH . '/assets/grid/available-configuration',
+        path: self::API_PATH . '/assets/grid/available-columns',
         operationId: 'asset_get_available_grid_configuration',
         description: 'asset_get_available_grid_configuration_description',
         summary: 'asset_get_available_grid_configuration_summary',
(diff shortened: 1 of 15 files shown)


src/Asset/Controller/Grid/Configuration/GetAvailableColumnsController.php
```diff
@@ -51,20 +51,20 @@ public function __construct(
...
-        '/assets/grid/available-configuration',
+        '/assets/grid/available-columns',
         name: 'pimcore_studio_api_get_asset_grid_available_configuration',
         methods: ['GET']
     )]
     #[IsGranted(UserPermissions::ASSETS->value)]
     #[Get(
-        path: self::API_PATH . '/assets/grid/available-configuration',
+        path: self::API_PATH . '/assets/grid/available-columns',
         operationId: 'asset_get_available_grid_configuration',
```
lukmzig:
```suggestion
        operationId: 'asset_get_available_grid_columns',
```

martineiber: Already Fixed it here :D
https://github.com/pimcore/studio-backend-bundle/pull/390/commits/21546aea5af75d7cbdc5d9e8fc7b8613e4ed4a2b
https://github.com/pimcore/studio-backend-bundle/pull/390/commits/e824b828233afe7b44caf644aba9685efe9fee16

lukmzig (approved):
🐧 :arrow_double_up:

src/Asset/Controller/Grid/Configuration/GetConfigurationController.php
```diff
@@ -88,15 +79,15 @@ public function getAssetGridConfiguration(
...
         #[MapQueryString] GridConfigurationIdParameter $configurationId = new GridConfigurationIdParameter()
     ): JsonResponse {
-        /**
-         * @todo: implement usage of $folderId and $configurationId
-         * If a configurationId is set, return the saved configuration for the folder.
-         * If no configurationId is set, return the default configuration.
-         */
-        $columns = $this->gridConfigurationService->getDefaultAssetGridConfiguration();
+        if ($configurationId->getConfigurationId() === null) {
+            $configuration = $this->gridConfigurationService->getDefaultAssetGridConfiguration();
```
mattamon: Hmm I am not sure if the controller should decide about this logic. Seems like this construct could also be integrated into the getAssetGridConfig, but this is just a feeling. Otherwise it is fine as it is.

mattamon (approved):
🐧 LGTM, just maybe think about the controller logic.

src/Asset/Controller/Grid/Configuration/GetConfigurationController.php
martineiber: I was also thinking about putting it into getAssetGridConfig or here into the controller. I went with the controller approach because  i think you can see it better what is going on. And since it is just a switch to different services i thought it would be okay todo it here. 

What does @lukmzig think about that?

lukmzig: I would say just to keep the idea that controller should not contain actual logic, we could move it to the getAssetGridConfig.

martineiber: Works for me! I will change it  :)

mattamon (approved):
🐈‍⬛ oh look, kitty is helping out to say now it is purrrrrfect.
🐧 approves too

PR #392, opened 2024-08-28
[Versions] Detail Endpoint Improvements

### LINKED ISSUE #391: [Versions] Detail Endpoint Improvements
* The version detail endpoint for documents currently only contains the filename but not the other attributes like creation date, mime type, filesize etc. Why?
image

* Would it be possible to add the asset type of the version to the endpoint (so image, document etc.)?

lukmzig:
Resolves #391 

## Additional info
Unify hydrator for asset versions


mattamon (approved):
🐧 good good

PR #404, opened 2024-09-02
Apply some CS Fixer Changes

martineiber:
Apply some CS-Fixer changes otherwise the other PR's will contain all those changes.


mattamon (approved):
🦭 look who is back from vacation! 

🦭 approves

PR #406, opened 2024-09-03
Problem when requesting 2 version preview image streams from the same asset ID in parallel

### LINKED ISSUE #393: Problem when requesting 2 version preview image streams from the same asset ID in parallel
When two version preview images with different binary content but the same asset image ID are requested in parallel or within a very short amount of time both return the same binary content:

image

This results in the problem that sometimes in the version comparison both images look the same allthough the binary content is different.

I was able to "fix" it by adding this line:
image

So it seems that there are some kind of race condition in the filesystem as the same filenames are used.

lukmzig:
Resolves #393

### DIFF
--- a/src/Version/Service/VersionBinaryService.php
+++ b/src/Version/Service/VersionBinaryService.php
@@ -87,14 +87,17 @@ final readonly class VersionBinaryService implements VersionBinaryServiceInterfa
         if (!$image instanceof Image) {
             throw new InvalidElementTypeException($image->getType());
         }
-        $thumbnail = $this->getImageThumbnail($image);
-
-        return $this->getStreamedResponse(
+        $originalName = $image->getFilename();
+        $thumbnail = $this->getImageThumbnail($image, $originalName);
+        $response = $this->getStreamedResponse(
             $thumbnail,
             HttpResponseHeaders::INLINE_TYPE->value,
             [],
             $thumbnail->getFileSize()
         );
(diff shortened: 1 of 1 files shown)


src/Version/Service/VersionBinaryService.php
```diff
@@ -127,4 +121,18 @@ public function streamPdfPreview(
 
         return $this->documentService->getPreviewStream($document);
     }
+
+    private function getImageThumbnail(Image $image): ThumbnailInterface
+    {
+        $image->setFilename(uniqid('', true));
```
mattamon: should the filename maybe the the original filename + unique?

markus-moser: And maybe we should reset the filename to the original one afterwards so that calling this service does not change the state of the asset instance?

lukmzig: I could not reproduce the issue ...but is it caused because of the response headers or because of the asset instance filename?

markus-moser: Because of the asset instance filename - also with disabled browser cache the same happens. I agree that it's not easy to reproduce without frontend.

lukmzig: @markus-moser can you please test if it works now?

markus-moser (approved):
@lukmzig I can confirm that this fixes the problem - now it works like it should!

PR #416, opened 2024-09-10
[Improvement] Created new schedule for update if no id is sent

### LINKED ISSUE #408: [Schedules| Allow creating new schedules in one go
The schedules list needs to be saved always alltogether either when saving the element or when pressing "Save scheduled tasks". Therefore it is hard to manage to create individual requests for each single row. 

The `/studio/api/schedules/{elementType}/{id}` `PUT` call allows updating of a whole list of schedule entries. It also manages deletions perfectly. So the only case which currently does not work is adding new schedules. 

Would it be possible to add this functionality and allow also the creation of new schedule entries (e.g. by leaving out the ID or as null value)?

mattamon:
Resolves #408 

## Additional info
If no id is sent in the update for task, a new one will be created.

### DIFF
--- a/src/Schedule/Repository/ScheduleRepository.php
+++ b/src/Schedule/Repository/ScheduleRepository.php
@@ -106,10 +106,11 @@ final readonly class ScheduleRepository implements ScheduleRepositoryInterface
 
         $currentTasks = [];
         foreach ($schedules as $schedule) {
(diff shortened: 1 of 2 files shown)


lukmzig (approved):

lukmzig (approved):
:giraffe:

PR #420, opened 2024-09-12
[User Management] List all available users

### LINKED ISSUE #398: [User Management] List all available users
We need an Endpoint which lists all available users. This Endpoint will then be used in different sections of the Studio UI. 
e.g. in the Grid configuration to share it with different users.

martineiber:
Resolves #398


mattamon (approved):
🦒 is looking from above on your code. 🦒 almost cannot read it but only almost.
It approves your code but finds it funny that you have to call `getUsers()` on the userListing and load is not enough.

PR #429, opened 2024-09-19
[Grid] Add PQLQuery Filter

### LINKED ISSUE #394: [Grid] Add PQLQuery Filter
image

martineiber:
Resolves #394 

## Additional info
Some minor Refactoring. Now it is possible to have a filter without a key, like `tag`, and `pql`. Those will be applied to the query and not to a column itself. For this we now have the `SimpleColumnFiltersParameterInterface`


lukmzig (approved):
:eagle:

PR #430, opened 2024-09-19
Introduce two new interfaces for DataIndex Query

### LINKED ISSUE #425: Introduce two new interfaces for DataIndex Query
We should add `AssetQueryInterface` and `DataObjectQueryInterface`

Those classes should implement those Interfaces.
https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/DataIndex/Query/AssetQuery.php
https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/DataIndex/Query/DataObjectQuery.php

And check in all filters the instanceof check and replace them with the interface.

martineiber:
Resolves #425

## Additional info
Add some missing unit tests for asset filters.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,3 +6,4 @@ services:
             PHP_IDE_CONFIG: serverName=localhost
         volumes:
             - .:/var/cli/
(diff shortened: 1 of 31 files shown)


lukmzig (approved):
:elephant:  LGTM :)

PR #438, opened 2024-09-25
Dependencies endpoints for data objects do not work correctly

### LINKED ISSUE #437: Dependencies endpoints for data objects do not work correctly
The following endpoint currently does not work:

`/studio/api/dependencies/data-object/82?page=1&pageSize=20&dependencyMode=required_by``

image

I guess this happens because of our various different spellings for the data object type (`object`, `data-object`, `dataObject`)

Maybe it helps using the `ElementType::fromShortValue()` method instead of the `ElementType::tryFrom()`?

lukmzig:
Resolves #437


martineiber (approved):
LGTM :bee:

PR #447, opened 2024-09-30
WIP: [Grid] [Asset] Refactor Data Source

### LINKED ISSUE #441: [Grid] [Asset] Refactor Data Source
Currently the grid for Assets uses the database as a datasource after filtering the Dataindex. 
But we should use only the GDI as a Datasource when loading multiple objects.  

Should also be consistent with Data Objects that we can use the same Resolver. 
Currently the resolver requires a `ElementIntrerface` with should then not be the case anymore.

martineiber:
Resolves #441


mattamon (approved):
LGTM 🐢

PR #470, opened 2024-10-08
[Grid] [DataObject] List available columns for class definition

### LINKED ISSUE #435: [Grid] [DataObject] List available columns for class definition
Part of https://github.com/pimcore/studio-ui-bundle/issues/296

martineiber:
Resolves #435

### DIFF
--- a/src/Grid/Column/Collector/DataObject/FieldDefinitionCollector.php
+++ b/src/Grid/Column/Collector/DataObject/FieldDefinitionCollector.php
@@ -49,8 +49,7 @@ final class FieldDefinitionCollector implements ColumnCollectorInterface, ClassI
     public function __construct(
         private readonly ClassDefinitionResolverInterface $classDefinitionResolver,
         private readonly DataObjectServiceResolverInterface $dataObjectServiceResolver,
-    )
-    {
+    ) {
     }
 
     public function getCollectorName(): string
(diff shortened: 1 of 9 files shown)


src/Grid/Column/Collector/DataObject/FieldDefinitionCollector.php
```diff
@@ -0,0 +1,179 @@
...
+    public function getColumnConfigurations(array $availableColumnDefinitions): array
+    {
+        try {
+            $classDefinition = $this->classDefinitionResolver->getById($this->getClassId());
+        } catch (Exception) {
+            $classDefinition = null;
+        }
+
+        if (!$classDefinition) {
+            throw new NotFoundException('Class definition', $this->getClassId());
```
mattamon, line 69: Can't we just throw the NotFoundException in the catch block above?

mattamon (changes requested):
Rest looks fantastic to me

src/Grid/Column/Collector/DataObject/FieldDefinitionCollector.php
martineiber: Unfortunately not...getById can return null, otherwise we have to throw it twice.

mattamon (approved):
🦊 foxy says go

PR #471, opened 2024-10-08
[Task] If empty list, add default metadata

### LINKED ISSUE #464: Add default image asset metadata (title, alt, copyright) to endpoints
In the classic bundle the default metadata attributes (title, alt, copyright) are added to the UI in ExtJS when the list is empty:

https://github.com/pimcore/admin-ui-classic-bundle/blob/c68397e4e4fa83d6179e1a04b75fc28f2747a4ac/public/js/pimcore/asset/metadata/grid.js#L30

Can we add them to the API instead so that we do not need to handle this in the UI?

mattamon:
Resolves #464 

## Additional info
If list is empty add default meta data

### DIFF
--- a/src/Asset/Service/Data/CustomMetadataService.php
+++ b/src/Asset/Service/Data/CustomMetadataService.php
@@ -35,6 +35,12 @@ final readonly class CustomMetadataService implements CustomMetadataServiceInter
 {
     use ElementProviderTrait;
 
+    private const DEFAULT_METADATA = [
+        ['name' => 'title', 'language' => '', 'type' => 'input', 'data' => ''],
+        ['name' => 'alt', 'language' => '', 'type' => 'input', 'data' => ''],
+        ['name' => 'copyright', 'language' => '', 'type' => 'input', 'data' => '']
+    ];
+
     public function __construct(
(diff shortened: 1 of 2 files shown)


src/Asset/Service/Data/CustomMetadataService.php
```diff
@@ -35,6 +35,12 @@
 {
     use ElementProviderTrait;
 
+    private const DEFAULT_METADATA = [
```
martineiber: Can we make that public?
Then  i could use is here as well, and we do not need do Hard Code this twice? 
https://github.com/pimcore/studio-backend-bundle/blob/91c9e46213fd2baa33005312620598f634324ff0/src/Grid/Column/Collector/Asset/MetadataCollector.php#L61

Or a getter would also work!

martineiber (changes requested):

src/Asset/Service/Data/CustomMetadataService.php
mattamon: Certainly, even moved them to the interface.

martineiber (approved):
:seal: loves default metadata.

PR #478, opened 2024-10-09
[Bug] Async processes do not have user in session

### LINKED ISSUE #476: [Bug]: Patch asynchronously does not work with usermodification
### Studio API version

1.x

### Steps to reproduce

Try to patch multiple assets via api asynchronously

### Actual Behavior

It will fail because not user in current session

### Expected Behavior

Works!

mattamon:
Resolves #476 

## Additional info
Async patches do not have user in session. To be able to set the userModification we need to use the user from the jobrun


martineiber (approved):
LGTM :seal:

PR #479, opened 2024-10-09
[Grid] [Asset] Save Configuration shared User and role can be empty.

### LINKED ISSUE #465: [Grid] [Asset] Save Configuration shared User and role can be empty.
It should be possible that shareUser and shareRole can be empty in the request

martineiber:
Resolves #465


mattamon (approved):
🐒  swings to approval

PR #491, opened 2024-10-15
[Task] Set pattern dynamically, add url_prefix check

### LINKED ISSUE #483: [Task] Make route for docs configurable.
Instead of hardcoded /studio/api/docs the studio part should be configurable.

mattamon:
Resolves #483 

## Additional info
Setting the firewall pattern dynamically for login


martineiber:
Can we then remove this? Or should we leave it?
https://github.com/pimcore/studio-backend-bundle/blob/f85ecc400df8a6427fc33df3dfa9dd91e26a8294/config/pimcore/firewall.yaml#L3

mattamon:
> Can we then remove this? Or should we leave it?
> 
> https://github.com/pimcore/studio-backend-bundle/blob/f85ecc400df8a6427fc33df3dfa9dd91e26a8294/config/pimcore/firewall.yaml#L3

I would leave it. I tried to set it via parameter but failed miserably.
It's kind of the default anyways.

martineiber (approved):
LGTM :duck:

PR #493, opened 2024-10-15
[Grid] [Asset] Save Configuration should return configuration schema

### LINKED ISSUE #484: [Grid] [Asset] Save Configuration should return id
When saving a new grid configuration the response should include the saved object with id.

martineiber:
Resolves #484

### DIFF
--- a/src/Asset/Controller/Grid/Configuration/SaveConfigurationController.php
+++ b/src/Asset/Controller/Grid/Configuration/SaveConfigurationController.php
@@ -16,12 +16,14 @@ declare(strict_types=1);
 
 namespace Pimcore\Bundle\StudioBackendBundle\Asset\Controller\Grid\Configuration;
 
+use OpenApi\Attributes\JsonContent;
 use OpenApi\Attributes\Post;
 use Pimcore\Bundle\StudioBackendBundle\Asset\Attribute\Request\Grid\SaveConfigurationRequestBody;
 use Pimcore\Bundle\StudioBackendBundle\Asset\MappedParameter\Grid\SaveConfigurationParameter;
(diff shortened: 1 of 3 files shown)


src/Asset/Controller/Grid/Configuration/SaveConfigurationController.php
```diff
@@ -72,8 +75,8 @@ public function __construct(
     public function saveAssetGridConfiguration(
         #[MapRequestPayload] SaveConfigurationParameter $saveConfigurationParameter
     ): Response {
-        $this->gridSaveConfigurationService->saveAssetGridConfiguration($saveConfigurationParameter);
+        $configuration = $this->gridSaveConfigurationService->saveAssetGridConfiguration($saveConfigurationParameter);
```
mattamon, line 78: Title and return does not match. Title says return id, but the save now returns the full configuration right?

martineiber: Yeah i was lazy to add more description :D
The main point was that @vin0401  needs an ID. We decided we will provide the same object as used in the listing of configurations. 

This is not the Full configuration this is a reduced version with ID, Name and Description

mattamon: Okay :D then please just update the title for the PR, so there is no confusion in the commit history :D

martineiber: JFYI: The full Configuration is under the DetailedConfiguration schema

mattamon (approved):
It got the looks.... of good!
🦊

PR #501, opened 2024-10-17
[Task] Change route from .json to /json

mattamon:
## Additional info
https://github.com/pimcore/demo-enterprise/blob/2024.x/.docker/nginx.conf#L113 -> the studio part can be removed.

### DIFF
--- a/doc/00_Installation.md
+++ b/doc/00_Installation.md
@@ -59,7 +59,7 @@ https://<your-pimcore-host>/pimcore-studio/api/docs
 
 You can also access the OpenApi documentation in JSON format by navigating to the following URL:
 ```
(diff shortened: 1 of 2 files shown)


martineiber (approved):
:duck: approves json as path

martineiber:
Just saw it.
We should also update this one to `docs/json`!
https://github.com/pimcore/studio-backend-bundle/blob/97188d332bc2d88b884fc383e4540cd9cdee87db/doc/00_Installation.md?plain=1#L62

mattamon:
> Just saw it. We should also update this one to `docs/json`!
> 
> https://github.com/pimcore/studio-backend-bundle/blob/97188d332bc2d88b884fc383e4540cd9cdee87db/doc/00_Installation.md?plain=1#L62

Done!

PR #502, opened 2024-10-17
[Workflow] Update workflows update and response schema

lukmzig:
- make submit schema more clear for workflow options
- objectLayout in allowed transitions can have multiple options


mattamon (approved):
🥇

PR #506, opened 2024-10-17
Update 00_Installation.md

Corepex:
Resolves #

## Additional info

Accoding to the latest changes i guess it should be `pimcore-studio` instead of `studio`


mattamon (approved):
Good catch. Thanks!

PR #508, opened 2024-10-17
[User] Colliding user routes for permissions and detail

### LINKED ISSUE #507: User permissions endpoint
The `/pimcore-studio/api/user/available-permissions` lists all permissions of a user. Is it intended to work for the authenticated user?

At the moment it delivers an error: `Pimcore\\Bundle\\StudioBackendBundle\\User\\Controller\\GetUserController::getUserById(): Argument #1 ($id) must be of type int, string given, called in /var/www/html/vendor/symfony/http-kernel/HttpKernel.php on line 181`.

We would need this endpoint exactly like this for the authenticated user for our permission handling. Could you please check?

lukmzig:
Resolves #507

### DIFF
--- a/src/User/Controller/GetUserController.php
+++ b/src/User/Controller/GetUserController.php
@@ -52,7 +52,7 @@ final class GetUserController extends AbstractApiController
     /**
      * @throws NotFoundException|DatabaseException
      */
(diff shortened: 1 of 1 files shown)


martineiber (approved):
LGTM :seal:

PR #510, opened 2024-10-21
Enable syncronus processing for creating a folder

### LINKED ISSUE #515: Creating Folder is handled async
The `/elements/asset/folder` endpoint is handled async since the request, immediately after the creation, doesn't return the new folder. Please change that so that we get the result immediately after the request is finished

mattamon:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/515


Corepex:
Resolves: #515

Corepex (approved):
Thanks for the quick fix - may the force be with you

PR #518, opened 2024-10-22
[Assets][Image] Add endpoint to return image preview based on studio config

### LINKED ISSUE #458: Missing functionalities regarding asset preview
We have some special cases regarding the asset preview which are currently missing:

* Image preview thumbnail with a higher resolution
Currently we only have the system thumbnail definition like in the classic bundle. This is fine for most cases but for the actual detail view of an asset we would need a higher resolution because our new design offers more space for the actual image + a zoom button.

* Video thumbnail generation loading state/progress
In the video preview we currently show the original asset video. We would need to show a video thumbnail there. In the classic bundle we have some kind of loading spinner while this preview is generated (can take a while...). We have already some endpoints in that direction but I'm not sure if something is missing or high exactly we need to call the endpoints to achieve the same.

* PDF preview javascript removal queue with loading state/progress
This is a similar case like the video thumbnail generation. In the classic bundle we have a messenger worker which removes javascript from PDF previews (for security reasons...). While the queue is running the classic bundle shows a loading indicator. We need the same in studio UI.

* Some months ago you mentioned that you find some special functionallities in the classic bundle regarding the asset preview. Where does the listed topics above or do you know about other special cases?

Could you please tell me who I should contact? I'd say it's better if we do a short call on those topics.

lukmzig:
Resolves #458 

### Additional info
Added a new endpoint to return image preview stream based on the pre-defined config

### DIFF
--- a/src/Asset/Controller/Image/PreviewStreamController.php
+++ b/src/Asset/Controller/Image/PreviewStreamController.php
@@ -1,0 +1,96 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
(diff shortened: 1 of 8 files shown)


mattamon (approved):
LGTM 🐺

PR #519, opened 2024-10-23
[Grid] [Resolver] Asset Documnet and DataObject Resolver for Metadata should not throw a NotFoundException

### LINKED ISSUE #517: [Grid] [Resolver] Asset Documnet and DataObject Resolver for Metadata should not throw a NotFoundException
Should return null if element not found it could be just that the user do not have the permission to see the Element

martineiber:
Resolves #517

### DIFF
--- a/src/Grid/Column/Resolver/Metadata/DocumentResolver.php
+++ b/src/Grid/Column/Resolver/Metadata/DocumentResolver.php
@@ -49,7 +49,7 @@ final class DocumentResolver implements ColumnResolverInterface
         }
 
         try {
-            $documentIndex = $this->documentService->getDocument(
+            $relatedDocument = $this->documentService->getDocument(
                 reset($document['document'])
             );
         } catch (NotFoundException) {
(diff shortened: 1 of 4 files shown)


src/Grid/Column/Resolver/Metadata/DocumentResolver.php
```diff
@@ -47,9 +48,13 @@ public function resolve(Column $column, ElementInterface $element): ColumnData
             return $this->getColumnData($column, null);
         }
 
-        $documentIndex = $this->documentService->getDocument(
-            reset($document['document'])
-        );
+        try {
+            $documentIndex = $this->documentService->getDocument(
```
lukmzig: should we not call this $relatedDocument just for consistency? :)

lukmzig (approved):

src/Grid/Column/Resolver/Metadata/DocumentResolver.php
martineiber: Yeah sure makes sense :+1:

mattamon (approved):
🐺

PR #520, opened 2024-10-28
[DataObject]: Add endpoint to list data object layouts

lukmzig:
Part of  https://github.com/pimcore/studio-ui-bundle/issues/265

## Additional info
- Added endpoint to list layouts for data objects
- properties might need to be adjusted based on the FE feedback
- custom layouts need to be implemented still and considered before returning final response


martineiber (approved):
LGTM :seal:

PR #525, opened 2024-10-31
[DataObject]: Fix data object layout endpoint when layout value in panel is null

markus-moser:
Unfortunately the getter in the Pimcore core has a bug. It's return type is not nullable allthough the value can get null. I therefore access the property directly for now:

image

https://github.com/pimcore/pimcore/blob/11.x/models/DataObject/ClassDefinition/Layout/Panel.php#L38


markus-moser:
Ahh, I see - you fixed it in the core:
https://github.com/pimcore/pimcore/pull/17780

PR #526, opened 2024-11-04
[Task] Add admin flag to user schema

### LINKED ISSUE #522: [User Management] User get by ID call doesn't return admin true/false
Currently it's possible to set the `admin` flag in the `POST` request but in the `GET` it seems to be missing:

Image

mattamon:
Resolves #522 

## Additional info
Add missing admin flag to user schema


martineiber (approved):
LGTM :seal:

PR #530, opened 2024-11-05
[Tags] Followup: Batch methods

### LINKED ISSUE #257: [Task] Tags followup: Batch methods + Mercure events
Still open is the batch handling of the tag assignment (maybe use another approach -> assign to childs instead of sending all ids)

Mercure notification should be also done as soon as batch assignment is ready.

-> check permission handling especially with batch operations.

lukmzig:
Resolves #257

## Additional info
- Merge endpoints and provide parameter to add or re-assign tags
- Use GEE and Mercure for processing


mattamon (approved):
🦥 Lazy approve

PR #532, opened 2024-11-05
[DataObject]: Apply dynamic data to data object layout endpoint

### LINKED ISSUE #528: [DataObject]: Apply dynamic data to data object layout endpoint
The layout endpoint (e.g. `/pimcore-studio/api/data-objects/1180/layout`) currently does not apply dynamic data like for example the select options if a select option provider is configured.

In the classic bundle the result looks like this:
Image

Whereas in Studio like this:
Image

Same applies also to e.g. the language select data type.

I guess it's all about respecting the `enrichLayoutDefinition` methods:
https://github.com/pimcore/pimcore/blob/322366f2d0d7613a0ce7ac796092f733b165008f/models/DataObject/ClassDefinition/Data/Select.php#L270

See:
https://github.com/pimcore/pimcore/blob/322366f2d0d7613a0ce7ac796092f733b165008f/models/DataObject/Service.php#L1347

lukmzig:
Resolves #528 

## Additional info
- add user for search query for proper permissions check

### DIFF
--- a/src/Asset/Service/AssetService.php
+++ b/src/Asset/Service/AssetService.php
@@ -86,6 +86,7 @@ final readonly class AssetService implements AssetServiceInterface
         );
 
         $assetQuery->orderByPath('asc');
+        $assetQuery->setUser($this->securityService->getCurrentUser());
 
         $result = $this->assetSearchService->searchAssets($assetQuery);
 
--- a/src/DataObject/Service/DataObjectService.php
+++ b/src/DataObject/Service/DataObjectService.php
@@ -129,6 +129,7 @@ final readonly class DataObjectService implements DataObjectServiceInterface
             $parameters,
             ElementTypes::TYPE_DATA_OBJECT
         );
+        $query->setUser($this->securityService->getCurrentUser());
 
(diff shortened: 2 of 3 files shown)


src/Asset/Service/AssetService.php
```diff
@@ -86,6 +86,7 @@ public function getAssets(ElementParameters $parameters): Collection
         );
 
         $assetQuery->orderByPath('asc');
+        $assetQuery->setUser($this->securityService->getCurrentUser());
```
mattamon, line 89: Not sure about this one, since I think we use the getAssets also for async calls, where we do not have a user no?

lukmzig: I did not find any usage of this in the async call, only in the tree endpoint. I think we query DB for async calls?

mattamon: > I did not find any usage of this in the async call, only in the tree endpoint. I think we query DB for async calls?

Hmm yeah not sure, I thought maybe in the handlers somewhere :) but if that is the case we can go ahead and merge it

mattamon (approved):
🇮🇹 approved

PR #533, opened 2024-11-05
[Task] Change permission to PIMCORE_USER

### LINKED ISSUE #531: `available-permissions` endpoint should be callable for every user
The `/pimcore-studio/api/user/available-permissions` endpoint should be callable without any extra permissions.

Api Link: http://localhost/pimcore-studio/api/docs#/User%20Management/user_get_available_permissions

Image

mattamon:
Resolves #531 

## Additional info
Changed permission to pimcore_user instead of users permission


martineiber (approved):
LGTM :seal:

PR #542, opened 2024-11-06
[Task] Add listener to enable options like in Classic UI

### LINKED ISSUE #534: [DataObject]: Apply dynamic data for user type to layout endpoint
This is a follow-up issue to #528. Currently it does not work for the user datatype. There we have the situation that it's not built by using the `enrichLayout` methods. I guess because this mechanism with the layout enrichment was invented later then the user datatype.

https://github.com/pimcore/pimcore/blob/11.x/models/DataObject/ClassDefinition/Data/User.php#L42

The easist solution would be to set the admin mode for the studio backend endpoints:
https://github.com/pimcore/pimcore/blob/11.x/lib/Pimcore.php#L53

Otherwise we could also refactor the data type in the core to use the `enrichLayout` approach. But I guess we will also have other situations where the admin mode is relevant.

lukmzig:
Resolves #534 

## Additional info
Add event listener which enables options on requests like in classic UI. See: https://github.com/pimcore/pimcore/blob/11.x/bundles/CoreBundle/src/EventListener/PimcoreContextListener.php#L81

### DIFF
--- a/config/event_subscribers.yaml
+++ b/config/event_subscribers.yaml
@@ -9,6 +9,10 @@ services:
     tags: [ 'kernel.event_subscriber' ]
     arguments: ['%pimcore_studio_backend.url_prefix%']
 
+  Pimcore\Bundle\StudioBackendBundle\EventSubscriber\ApiContextSubscriber:
+    tags: [ 'kernel.event_subscriber' ]
+    arguments: ['%pimcore_studio_backend.url_prefix%']
+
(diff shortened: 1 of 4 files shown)


mattamon (approved):
As max verstappen said after winning in Brazil. Simply lovely 🏎️

dpfaffenbauer:
@mattamon amazing race that was ;)

PR #545, opened 2024-11-08
Extend current-user-information endpoint

### LINKED ISSUE #538: Extend `current-user-information` endpoint
Related to: https://github.com/pimcore/studio-backend-bundle/issues/539

As discussed, we need to adjust the `/pimcore-studio/api/user/current-user-information` endpoint.

Please add the following fields.

- `permissions` (string[]) - an array of strings with all permissions the user is granted
- `admin` (bool) - or `isAdmin` if it fits better ...

IMHO, we can remove the `roles` since there is no actual use case on the front end - the `isAdmin` flag will replace it.

Please also adjust the response of `/pimcore-studio/api/login`  - it should return the same as `/pimcore-studio/api/user/current-user-information`

martineiber:
Resolves #538


src/Authorization/Controller/LoginController.php
```diff
@@ -50,11 +51,14 @@ final class LoginController extends AbstractApiController
...
-    {
-        return $this->jsonResponse([
-            'username' => $user->getUserIdentifier(),
-            'roles' => $user->getRoles(),
-        ]);
+    public function login(
+        #[CurrentUser] User $user,
+        UserInformationHydratorInterface $userInformationHydrator
+    ): JsonResponse {
+        $userInformation = $userInformationHydrator->hydrate(
```
lukmzig: What do you think if we would move this into the service? I guess we also should dispatch an event again if someone wants to add some additional data? Or did we decide that we will not do so for the user information?

martineiber: Good Point with the Event. At first i thought i will not introduce an Service since it will just call the Hydrator. 
But with the Event it totally make sense. 

I will add this.

mattamon: Also please keep in mind our blueprint, which states that the controller delegates to the service and the service then has the hydrator. :)

martineiber: Done :)

mattamon (approved):
🦄 LGTM

PR #548, opened 2024-11-11
Add available admin languages

### LINKED ISSUE #547: Available admin languages endoint
We would need (e.g. for our user management) an endpoint with the availalbe admin languages. It should provide the same data like the `/admin/settings/get-available-admin-languages` endpoint from the classic bundle. We could maybe think about adding this to the `/pimcore-studio/api/settings` endpoint or introduce a new one.

mattamon:
Resolves #547


lukmzig (approved):
LGTM :abcd:

Issue #546, opened 2024-11-11
[User Management] API improvements


markus-moser:
- [x] Add a fulltext search query parameter to the `/api/users/tree` endpoint. We need this for our search field in the frontend.
Search for `firstname, lastname, email, id` like https://github.com/pimcore/admin-ui-classic-bundle/blob/1.x/src/Controller/Admin/UserController.php#L932
 Will be a new user/search endpoint.

- [x] Add the user references/dependencies to the detail endpoint. We need this for our "This user is referenced in the following objects" tab:
![Image](https://github.com/user-attachments/assets/7fca3c81-2ee9-43bd-84ba-587e60050c20)
It should be the same data like in the `objectDependencies` attribute of our user detail endpoint from the classic bundle (`hasHidden` flag + `dependencies`):
![Image](https://github.com/user-attachments/assets/33b10893-3656-482a-9a90-47251723dd87)

- [x] Updating the `keyBindings` currently produces an error. We send the data the same way like we receive it from the get endpoint:
![Image](https://github.com/user-attachments/assets/935b25fb-f6fa-4777-9b98-f9246b2376cf)

- [x] For a user without permissions (directly after creating a new one) we currently get a permission array with one single empty string as entry. Please change it to an empty array:
![Image](https://github.com/user-attachments/assets/0e465768-20ca-4741-aeb4-f2ff130174a6)

martineiber:
Key Binding is now also fixed by #722

PR #552, opened 2024-11-14
[User Management] Add user Search

martineiber:
part of #546

### DIFF
--- a/src/User/Repository/UserRepository.php
+++ b/src/User/Repository/UserRepository.php
@@ -146,6 +146,7 @@ final readonly class UserRepository implements UserRepositoryInterface
     {
         $list = [];
         $q = '%' . $searchQuery . '%';
+
         try {
             $userListing = new UserListing();
             $userListing->setCondition(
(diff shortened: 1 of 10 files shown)


src/User/Repository/UserRepository.php
```diff
@@ -136,4 +136,36 @@ public function getUsers(): array
...
+    public function searchUser(string $searchQuery): array
+    {
+        $list = [];
+        $q = '%' . $searchQuery . '%';
+
+        try {
+            $userListing = new UserListing();
+            $userListing->setCondition(
+                'name LIKE ? OR firstname LIKE ? OR lastname LIKE ? OR email LIKE ? or id = ?',
+                [$q, $q, $q, $q, (int)$searchQuery]
```
mattamon, line 154: Hmm, is it really necessary that you could search for the ID? 
Like then you can basically found out every id of every user no just by "brute forcing" the search?

martineiber: Yes it is, since it is the Requirement for the search, that you can search for the ID as well. 

Well to get all ID's of all users you can just list all users, that would be easier  :)

mattamon (approved):
LGTM 🐭

PR #580, opened 2024-11-27
Update to enable and fix tests.

herbertroth:
Resolves #


src/DataObject/Attribute/Parameter/Query/ClassNameParameter.php
```diff
@@ -37,7 +37,7 @@ public function __construct()
             schema: new Schema(
                 type: 'string',
                 enum:  array_map(static fn (ClassDefinition $def) => $def->getName(), $classDefinitions->load()),
-                example: null
+                example: 'string'
```
mattamon:
```suggestion
                example: '--'
```
I guess if the example is not in the enum, it will resort to null right?
Maybe we should use the '--' here since that how it is shown in the docs
image

Maybe removing the example all together does work?

src/DataObject/Controller/TreeController.php
```diff
@@ -78,7 +78,7 @@ public function __construct(
     #[PageSizeParameter]
     #[ParentIdParameter(
         description: 'Filter data objects by parent id.',
-        example: null
+        example: 12
```
mattamon:
```suggestion
        example: 1
```
Let's rather use 1 here since it would be the root folder.

src/Asset/Controller/TreeController.php
```diff
@@ -76,7 +76,7 @@ public function __construct(
     #[PageSizeParameter]
     #[ParentIdParameter(
         description: 'Filter assets by parent id.',
-        example: null,
+        example: 83,
```
mattamon:
```suggestion
        example: 1,
```
Let's rather use 1 since it is the root folder

src/OpenApi/Config/Spec.php
```diff
@@ -28,12 +30,21 @@
...
+)]
+
 #[License(
     name: 'GNU General Public License version 3 & Pimcore Commercial License',
-    identifier: 'GPLv3 & PCL',
     url: 'https://www.pimcore.org/license'
 )]
 
+#[Server(
+    url: 'http://localhost:6001',
```
mattamon: We should not set this as a fixed value.
Like we discussed, we can inject it in the OpenApi Service. 
I would remove it for now and then add another PR where you can configure this.

mattamon (changes requested):

PR #583, opened 2024-11-27
[Improvement]: Ignore undefined data object adapters for now 

lukmzig:

Ignore undefined data object adapters for now as not all field types from all bundles are implemented yet


mattamon (approved):
LGTM

martineiber (approved):
LGTM :seal:

lukmzig:
@martineiber @mattamon 
I did some additional changes. The values for setter should be rather skipped for undefined field types instead of setting them to null. We also then need a special check for the encrypted fields unfortunately as those are having "delegated" field type.

Issue #566, opened 2024-11-22
[Data Object Editor] Endpoints for editor


markus-moser:
*Top prio:*
 
* [x] Regular `GET` endpoint for the data object data (based on field definition adapters and `normalize()` methods of the field defintions)

* [x] In the classic bundle `/admin/object/get` we also have additional relevant data which should be added to our new get endpoint (let's concentrate on the most important ones for now and see what we need in addition to that as soon as things become relevant):
  * allowInheritance
  * allowVariants
  * showVariants
  * hasPreview

* A pendant for the following endpoints from classic bundle. Please take a look at the actions how they are implemented, they should offer the same possibilities and provide the same data:

    * [x] `/admin/class/fieldcollection-tree?_dc=1732263392203&allowedTypes=NewsCars%2CNewsLinks%2CNewsText&object_id=773&field_name=content&layoutId=0&forObjectEditor=1`(we only need the `forObjectEdit` relevant infos without custom layouts for now)

    * [x] /admin/class/objectbrick-tree?_dc=1732263612453&class_id=AP&object_id=408&field_name=saleInformation&layoutId=0&forObjectEditor=1 (we only need the `forObjectEdit` relevant infos without custom layouts for now)

    * [x] `/admin/quantity-value/unit-list`.
    * [x] `/admin/settings/get-available-sites`  (we can skip the usesless "dao" attribute in the response)
    * [x] `/admin/quantity-value/convert`
    * [x] `/admin/quantity-value/convert-all`
    * [x] `pimcore_admin_dataobject_dataobject_getSelectOptions` route

*Additional topics:*

* Please analyze the following routes of the classic bundle - do our new pendants of the backend API offer the same possibilites or are relevant topics missing?
     * [x] `pimcore_admin_asset_getimagethumbnail`
     * [x] ~~`pimcore_admin_asset_download`~~
     * [x] `pimcore_admin_asset_getvideothumbnail`

* For relation data types, there are several more routes involved. Didn't analyze them in detail right now, we need to find the right approach to solve those topics together:

   * [x] `pimcore_admin_dataobject_dataobject_add`
   * [ ] `pimcore_admin_dataobject_dataobjecthelper_loadobjectdata`
   * [x] `pimcore_admin_dataobject_dataobject_get`
   * [x] `pimcore_admin_element_lockelement`
   * [x] `pimcore_admin_element_getnicepath`
   * [x] `pimcore_admin_element_getsubtype`
  

[...]

PR #591, opened 2024-11-28
[Data Object Editor]: Resolve normalizer issues

### LINKED ISSUE #584: [Data Object Editor] Enrich consent data type with note text
In the classic bundle we get it like this:

Image

The `normalize()` method skips the `noteContent`, therefore we need to override the default implementation.

part of #566

### LINKED ISSUE #585: [Data Object Editor] ManyToManyRelation type use real full path
Currently `$reference->getFullPath()` is used. Please replace it with `$reference->getRealFullPath()`. The difference is that with `getFullPath` e.g. spaces are URL encoded etc. We need to original full path.

part of #566

### LINKED ISSUE #587: [Data Object Editor] Block data type broken
Currently it looks like this:

Image

Part of #566

### LINKED ISSUE #588: [Data Object Editor] Structure for field collections
The current structure for the field collections is a bit limiting when it comes to new information we want to add to the field collections:

Image

Instead we should nest the data in a own property to avoid naming conflicts:

```
FieldCollectionExample: [
    0: {
        type: 'MyCollction',
        data: {
            field1: 'value',
            field2: 'value'
        }
    }
]
```

### LINKED ISSUE #594: [Data Object Editor] Data format of RGBA color data type
In the classic bundle we get a hex code:

Image

The `normalize()` method produces an object like this:

Image

Please change it to the hex code, this makes it easier to be compatible with ant design's color picker.

lukmzig:
Resolves #584 
Resolves #585 
Resolves #587 
Resolves #588 
Resolves #594

part of https://github.com/pimcore/studio-backend-bundle/issues/566

### DIFF
--- a/src/DataObject/Data/Adapter/AdvancedManyToManyObjectRelationAdapter.php
+++ b/src/DataObject/Data/Adapter/AdvancedManyToManyObjectRelationAdapter.php
@@ -17,7 +17,7 @@ declare(strict_types=1);
 namespace Pimcore\Bundle\StudioBackendBundle\DataObject\Data\Adapter;
 
 use Pimcore\Bundle\StaticResolverBundle\Models\DataObject\ConcreteObjectResolverInterface;
-use Pimcore\Bundle\StudioBackendBundle\DataObject\Data\FieldContextData;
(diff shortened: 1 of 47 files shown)


mattamon (approved):
It got the looks

PR #614, opened 2024-12-03
[Feature][Asset] Stream image thumbnail action

### LINKED ISSUE #607: [Asset] Stream image thumbnail action
For our data object editor and other functionallities we need the possibility to render/stream a custom thumbnail definition based on an API endpoint / URL.

It can work similar to the already existing `/pimcore-studio/api/assets/{id}/image/stream/preview` endpoint but with the possibility to customize the thumbnail format.

We need the same approach like in the classic bundle in this example:

`/admin/asset/get-image-thumbnail?id=67&width=590&height=400&contain=true`

Here it automatically maps all GET parameters to the thumbnail definition:

https://github.com/pimcore/admin-ui-classic-bundle/blob/89c871372cda984f86318eaefe2e2a0ea5ff4f5e/src/Controller/Admin/Asset/AssetController.php#L1183

For sure it would be ok to allow only certain parameters. Let's start with the following and see if we need more:
* width
* height
* contain
* frame
* cover
* format

Related to #566

mcop1:
Resolves #607

### DIFF
--- a/src/Asset/MappedParameter/ImageDownloadConfigParameter.php
+++ b/src/Asset/MappedParameter/ImageDownloadConfigParameter.php
@@ -102,30 +102,21 @@ final readonly class ImageDownloadConfigParameter
 
     public function getCoverTransformation(): array
     {
-        return [
-            'width' => $this->getWidth(),
-            'height' => $this->getHeight(),
-            'forceResize' => $this->getForceResize(),
+        return
+            [
+            ... $this->getBaseTransformationValues(),
             'positioning' => $this->getPositioning(),
         ];
     }
 
     public function getFrameTransformation(): array
     {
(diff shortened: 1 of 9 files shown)


src/Asset/MappedParameter/ImageDownloadConfigParameter.php
```diff
@@ -72,4 +98,70 @@ public function getDpi(): ?int
     {
         return $this->dpi;
     }
+
+    public function getCoverTransformation(): array
+    {
+        return
+            [
+            ... $this->getBaseTransformationValues(),
```
mattamon, line 105: Could we please update the formatting here? :)

lukmzig (approved):

PR #615, opened 2024-12-04
[Task] Add roave/security-advisories

martineiber:
See Title


mattamon (approved):
Security problems solved with more security updates!

PR #619, opened 2024-12-04
[Data Object Editor] Quantity value endpoints tag handling

### LINKED ISSUE #598: [Data Object Editor] Quantity value endpoints tag handling
* Please use a different tag then "Class Definition" - it should be a separate tag specialized for the units
* The convert and convert all endpoints should not invalidate any tag as they do not change data. Otherwise RTK query automatically refreshes the local unit list when calling these methods.

lukmzig:
Resolves #598 

## Additional info
- changed tag from class definitions to units
- changed requests from POST to GET


martineiber (approved):
LGTM :seal:

PR #621, opened 2024-12-09
[Data Object Tree] Change className property to string instead of enum

### LINKED ISSUE #620: [Data Object Tree] Change className property to string instead of enum
At the moment the `className` property is an enum with the available class definitions. For us in the frontend it's hard to work with this as it depends on the installation. With typescript we need to define the types but in that case it's not possible to define the correct type as it is flexible.

For example in our generated schema we then have the problem that the class names are added there:
https://github.com/pimcore/studio-ui-bundle/blob/8ccfb1a80c7deb2d1418e9797bf5dcf0d5c0c641/assets/js/src/core/modules/data-object/data-object-api-slice.gen.ts#L225

Could you therefore please change it to a string? On other locations we already have a string for the classId/className.

lukmzig:
Resolves #620


mattamon (approved):
LGTM

PR #623, opened 2024-12-09
[Workflows] Schema

### LINKED ISSUE #555: Object Workflow
Image
I am getting an error response when trying to submit a workflow action for objects. 

here my api call with parameters:
Image

The class name of the object is "Car":
Image

Which is the same class name that the workflow is specified for:
Image

When I remove the condition of the workflow I am getting the below error:
Image
Which may suggest there is a problem with getting Workflows for objects?

Image
For consistency purposes we may want to use 'data-object' instead of 'object' here.

lukmzig:
Resolves #555 

## Additional info
- adapt schema to fit FE
- catch syntax error from symfony expression

### DIFF
--- a/src/Workflow/Schema/SubmitAction.php
+++ b/src/Workflow/Schema/SubmitAction.php
@@ -42,13 +42,13 @@ final readonly class SubmitAction
             example: WorkflowActionTypes::TRANSITION_ACTION
         )]
         private string $actionType,
-        #[Property(description: 'id of the element', type: 'integer', example: 50)]
+        #[Property(description: 'Id of the element', type: 'integer', example: 50)]
         private int $elementId,
-        #[Property(description: 'type of the element', type: 'string', example: ElementTypes::TYPE_DATA_OBJECT)]
+        #[Property(description: 'Type of the element', type: 'string', example: ElementTypes::TYPE_DATA_OBJECT)]
         private string $elementType,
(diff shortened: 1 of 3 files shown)


src/Workflow/Schema/SubmitAction.php
```diff
@@ -44,12 +44,12 @@ public function __construct(
...
-        #[Property(description: 'type of the element', type: 'string', example: ElementTypes::TYPE_OBJECT)]
+        #[Property(description: 'type of the element', type: 'string', example: ElementTypes::TYPE_DATA_OBJECT)]
         private string $elementType,
-        #[Property(description: 'name of the workflow', type: 'string', example: 'my_first_workflow')]
-        private string $workflowName,
-        #[Property(description: 'transition', type: 'string', example: 'start_workflow')]
-        private string $transition,
+        #[Property(description: 'id of the workflow', type: 'string', example: 'my_first_workflow')]
+        private string $workflowId,
+        #[Property(description: 'id of the transition', type: 'string', example: 'start_workflow')]
```
mattamon:
```suggestion
        #[Property(description: 'ID of the transition', type: 'string', example: 'start_workflow')]
```

lukmzig: Afaik we use id in the description of params ...
e.g
image

mattamon (approved):
🕺🏻 can't stop the feeling that this work flows

PR #626, opened 2024-12-10
[Asset Metadata] List asset predefined metadata

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/717

## Additional info
- add endpoint to list asset predefined metadata


lukmzig:
Not necessary as it should be already addressed by https://github.com/pimcore/studio-backend-bundle/pull/596

PR #632, opened 2024-12-11
[Asset Replace] Return original file name not the new one

### LINKED ISSUE #605: Wrong behaviour in `asset_replace` endpoint
Hi guys, 

I implemented the "upload new version" function in the frontend and used the following endpoint for that: 
http://localhost/pimcore-studio/api/docs#/Assets/asset_replace

I noticed that this endpoint changes the name of the whole asset. It should only patch the extension (if it is a different one)
soo `foobar.png` becomes `foobar.jpg`. 

Would be great if you could change that ...
If there are any questions let me know :)

lukmzig:
Resolves #605


mattamon (approved):
LGTM

PR #636, opened 2024-12-11
[Asset Replace] Return new file name as response

### LINKED ISSUE #633: Change response of `asset_replace`
Hi Guys, 

could you please change the response of `/pimcore-studio/api/assets/{id}/replace`. 
It would be great if the endpoint would return me the name of the new asset ... this would save us one extra request where we have to reload the tree to display the new name.

Thanks 🚀

lukmzig:
Resolves #633


martineiber (approved):
LGTM :+1:

PR #643, opened 2024-12-16
[Asset] Stream image thumbnail action improvements

### LINKED ISSUE #607: [Asset] Stream image thumbnail action

### LINKED ISSUE #624: [Asset] Stream image thumbnail action improvements
This is a follow up to #614 

I tested it with this call:

`/pimcore-studio/api/assets/{ID}/image/stream/custom?mimeType=JPEG&resizeMode=scaleByWidth&height=200&width=100&contain=true`

The pendant in the classic bundle would be:

`/admin/asset/get-image-thumbnail?id={ID}&width=200&height=100&contain=true`

* I received a strange result for a portrait image.
Studio:
Image
Classic Bundle:
Image

* Tested it also with an icon SVG:
Studio:
Image
Classic Bundle:
Image

* Not sure if my calls are wrong and the URL format needs to be different but at the moment it's strange that I need to combine the resizeMode with contain, cover and frame to create the URL. Maybe `contain`, `cover`and `frame` need to get resize modes instead? 

I tested with the following images:
[Archiv.zip](https://github.com/user-attachments/files/18065306/Archiv.zip)

mcop1:
Resolves #607 
Resolves #624 

## Additional info
ResizeMode is optional now.

### DIFF
--- a/src/Asset/Controller/Image/CustomStreamController.php
+++ b/src/Asset/Controller/Image/CustomStreamController.php
@@ -40,6 +40,8 @@ use Pimcore\Bundle\StudioBackendBundle\OpenApi\Attribute\Response\DefaultRespons
 use Pimcore\Bundle\StudioBackendBundle\OpenApi\Attribute\Response\SuccessResponse;
 use Pimcore\Bundle\StudioBackendBundle\OpenApi\Config\Tags;
 use Pimcore\Bundle\StudioBackendBundle\Security\Service\SecurityServiceInterface;
+use Pimcore\Bundle\StudioBackendBundle\Util\Constant\Asset\MimeTypes;
+use Pimcore\Bundle\StudioBackendBundle\Util\Constant\Asset\ResizeModes;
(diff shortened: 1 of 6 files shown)


lukmzig (approved):
LGTM

PR #647, opened 2024-12-18
[Data Objects]: Extend Patcher to use field adapters for editable data

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/300

## Additional info
- extend Patcher to update editable data - can be used for inline editing and detail editor


mcop1 (approved):
LGTM

PR #650, opened 2025-01-07
Added check if `dataUri` exists

Corepex:
Added check if `dataUri` is present

## Additional info
An error is thrown if you try to update an asset because `dataUri` is missing in the update body. 
Since it's not required (afaik), I would add a check if it's there. 

WDYT?


Corepex:
The issue seems to be introduced with https://github.com/pimcore/pimcore/pull/17941/files#diff-fd51b37126eb63374242a3788ce8cb52ae243560ab4e55e9a296ac0582a59e65

lukmzig (approved):
LGTM

PR #652, opened 2025-01-08
[Task] Use boolean arguments for mapped parameters directly

### LINKED ISSUE #649: [Video Assets] Image thumbnail aspectRatio=true seems to not work
The following example call:

`/pimcore-studio/api/assets/353/video/stream/image-thumbnail?width=500&height=500&aspectRatio=true` 

should respect the aspect ratio of the image thumbnail of a video. Currently this seems to not work:

Image

lukmzig:
Resolves #649 

## Additional info
Seems like boolean parameters are now supported directly


mattamon (approved):
True or false that is the question.
Boolean on my young padawan. ⚡

PR #654, opened 2025-01-08
[Bug] Fix object-brick structure

### LINKED ISSUE #653: [Data Object Editor] Structure of object bricks
The current structure of the object bricks seems broken:

Image

I would recommend to use the same structure like we use it for the field collections: https://github.com/pimcore/studio-backend-bundle/issues/588

mattamon:
Resolves #653 

#Addtional Infos
Looks like this now:
image


lukmzig (approved):
LGTM!

PR #667, opened 2025-01-09
Add permissions and locked status to grid items

### LINKED ISSUE #663: Add `permissions` to grid
Please always return the `permissions` when returning the grid config in the following endpoint `/pimcore-studio/api/assets/grid`

I would just add it on the same level as `colums`

mattamon:
Resolves #663 

## Additional info
Adding addtional infos to grid items

### DIFF
--- a/src/Grid/Attribute/Property/GridCollection.php
+++ b/src/Grid/Attribute/Property/GridCollection.php
@@ -30,8 +30,13 @@ final class GridCollection extends Property
     {
         parent::__construct(
             property: 'items',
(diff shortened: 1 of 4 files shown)


Corepex (changes requested):
I guess we also have to update the schema definition in `Grid/Attribute/Property/GridCollection.php` or?

Corepex (changes requested):
i would set `isLocked` and `permissions` required.
Please also add the `id` to the response (as required).

Corepex (approved):
awesome work thx :D

PR #679, opened 2025-01-13
[Improvement] Get fullpath for link

### LINKED ISSUE #678: [Data Object Editor] Path property in link type
Similar to #671 

We need the `path` (or `fullPath`) property also in the link data type. This is especially relevant when we have the linktype "internal" - so when we link to an Pimcore element. For the link type "direct" we could also fill up this property with the path from the "direct" property (but would also be ok if you leave it as null in that case).

mattamon:
Resolves #678 

## Additional info
Adding fullPath also to link element
Loading element if it is internal.
Using direct as fullpath if direct link


lukmzig (approved):
LGTM :giraffe:

PR #693, opened 2025-01-15
Add simple search endpoint

### LINKED ISSUE #656: [Task] Search endpoint via GDI functionality
https://github.com/pimcore/studio-ui-bundle/issues/245 -> Simple search for all element types

See current quick-search:
http://localhost/admin/bundle/search/search/quicksearch?query=test&page=1&start=0&limit=25

Image

Think of a concept how to use the global search also for all the subtypes without needing extra endpoints.

Use Fulltext Search from GDI. 

Fulltext Search needs to be weighted to deliver better results. - Needs to be decided by PO, PM

Concept -> Admin could use PQL to search for specific key value pairs

lukmzig:
Resolves #656 
Part of: https://github.com/pimcore/studio-ui-bundle/issues/245

## Additional info
Add endpoint to return list for Simple search using full text search of GDI

### DIFF
--- a/config/search.yaml
+++ b/config/search.yaml
@@ -1,0 +1,33 @@
+services:
+  _defaults:
+    autowire: true
+    autoconfigure: true
+    public: false
+
+  # controllers are imported separately to make sure they're public
+  # and have a tag that allows actions to type-hint services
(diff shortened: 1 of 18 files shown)


mattamon (approved):
Looks like the search for a search functionality is over

PR #708, opened 2025-01-20
[Data Object][Versions] Adjust response data format of data object versions

### LINKED ISSUE #689: [Data Object][Versions] Adjust response data format of data object versions
Adjust response of `/pimcore-studio/api/versions/{id}` for data objects to align with the detail endpoint. The data object attributes should have the same structure as detail endpoint (re-use data adapters)

lukmzig:
Resolves #689 

## Additional info
Data objects version endpoint needs to return data in same format as detail endpoint as the FE implementation differs from classic UI

### DIFF
--- a/src/Version/Schema/DataObjectVersion.php
+++ b/src/Version/Schema/DataObjectVersion.php
@@ -39,12 +39,8 @@ final class DataObjectVersion extends Element implements AdditionalAttributesInt
     public function __construct(
         #[Property(description: 'Key', type: 'string', example: 'Giulietta')]
         private readonly string $key,
-        #[Property(description: 'Class name', type: 'string', example: 'car')]
-        private readonly ?string $className = null,
         #[Property(description: 'Type', type: 'string', example: 'image')]
         private readonly string $type,
-        #[Property(description: 'Published', type: 'bool', example: false)]
-        private readonly ?bool $published = null,
         #[Property(description: 'Has children', type: 'bool', example: false)]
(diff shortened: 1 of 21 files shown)


src/Version/Schema/DataObjectVersion.php
```diff
@@ -18,41 +18,108 @@
...
+use Pimcore\Bundle\StudioBackendBundle\DataObject\Util\Trait\ClassDataTrait;
+use Pimcore\Bundle\StudioBackendBundle\Response\Element;
+use Pimcore\Bundle\StudioBackendBundle\Response\ElementIcon;
 use Pimcore\Bundle\StudioBackendBundle\Util\Schema\AdditionalAttributesInterface;
 use Pimcore\Bundle\StudioBackendBundle\Util\Trait\AdditionalAttributesTrait;
+use Pimcore\Bundle\StudioBackendBundle\Util\Trait\WorkflowAvailableTrait;
 
 #[Schema(
     title: 'DataObjectVersion',
     required: ['modificationDate', 'path', 'published'],
```
martineiber: We are missing some required fields here.

martineiber (changes requested):

src/Version/Schema/DataObjectVersion.php
lukmzig: Good point, thank you!

martineiber (approved):

PR #728, opened 2025-01-24
[Improvement] Data Objects - Custom Layouts

### LINKED ISSUE #658: [Task] Data Objects - Custom Layouts
We need to rethink the whole custom layout logic. We cannot use the same like in the admin-ui.

mcop1:
Resolves #658 

## Additional info
Depends on https://github.com/pimcore/static-resolver-bundle/pull/89
Depends on https://github.com/pimcore/static-resolver-bundle/pull/90


src/Asset/Service/DownloadService.php
```diff
@@ -134,4 +136,17 @@ public function downloadImageByThumbnail(
...
     }
+
+    public function downloadJSON(
+        string $json,
+        string $filename
+    ): JsonResponse {
+        $response = new JsonResponse(
+            $json
+        );
+        $response->headers->set('Content-Type', MimeTypes::JSON->value);
```
lukmzig: We can use ENUMS for header type here as well https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Util/Constant/HttpResponseHeaders.php

src/Class/Controller/CustomLayout/CollectionController.php
```diff
@@ -0,0 +1,87 @@
...
+
+    public function __construct(
+        SerializerInterface $serializer,
+        private readonly CustomLayoutServiceInterface $customLayoutService
+    ) {
+        parent::__construct($serializer);
+    }
+
+    /**
+     * @throws Exception|NotFoundException
```
lukmzig: Afaik we should try to catch and throw only studio exceptions from controller so we can control the default responses

src/Class/Controller/CustomLayout/CollectionController.php
```diff
@@ -0,0 +1,87 @@
...
+        example: 'CAR',
+        description: 'class_custom_layout_collection_data_object_class',
+        required: true
+    )]
+    #[SuccessResponse(
+        description: 'class_custom_layout_collection_success_response',
+        content: new CollectionJson(new GenericCollection(CustomLayoutCompact::class))
+    )]
+    #[DefaultResponses([
+        HttpResponseCodes::NOT_FOUND,
```
lukmzig: We need to probably add other default responses since we also throw Exception

src/Class/Controller/CustomLayout/CreateController.php
```diff
@@ -0,0 +1,85 @@
...
+        description: 'pimcore_studio_api_class_custom_layout_create_layout_id',
+        required: true
+    )]
+    #[CustomLayoutNewRequestBody]
+    #[SuccessResponse(
+        description: 'pimcore_studio_api_class_custom_layout_create_success_response',
+        content: new JsonContent(ref: CustomLayout::class, type: 'object')
+    )]
+    #[DefaultResponses([
+        HttpResponseCodes::UNAUTHORIZED,
```
lukmzig, line 76: Seems like there are some default responses missing

src/Class/Controller/CustomLayout/DeleteController.php
```diff
@@ -0,0 +1,74 @@
...
+        name: 'customLayoutId',
+        example: 'CarTodo',
+        description: 'pimcore_studio_api_class_custom_layout_delete_layout_id',
+        required: true
+    )]
+    #[SuccessResponse(
+        description: 'pimcore_studio_api_class_custom_layout_delete_success_response',
+    )]
+    #[DefaultResponses([
+        HttpResponseCodes::NOT_FOUND,
```
lukmzig, line 69: Are these correct? Seems like we are throwing `DefinitionWriteException` here

src/Class/Controller/CustomLayout/ExportController.php
```diff
@@ -0,0 +1,81 @@
...
+        description: 'pimcore_studio_api_class_custom_layout_export_layout_id',
+        required: true
+    )]
+    #[SuccessResponse(
+        description: 'pimcore_studio_api_class_custom_layout_export_success_response',
+        content: [new MediaType(MimeTypes::JSON->value)],
+        headers: [new ContentDisposition(fileName: 'custom_definition_CarToDo_export.json')]
+    )]
+    #[DefaultResponses([
+        HttpResponseCodes::NOT_FOUND,
```
lukmzig, line 75: Seems like there are some default responses missing

src/Class/Controller/CustomLayout/ImportController.php
```diff
@@ -0,0 +1,120 @@
...
+     * @throws ForbiddenException
+     * @throws NotFoundException
+     * @throws UserNotFoundException
+     */
+    #[Route(
+        '/class/custom-layout/import/{customLayoutId}',
+        name: 'pimcore_studio_api_class_custom_layout_import',
+        methods: ['POST']
+    )]
+    #[IsGranted(UserPermissions::ASSETS->value)]
```
lukmzig: It probably should be Data objects or?

lukmzig (changes requested):
In general it seems like that controllers are missing some default responses and maybe we can double-check the exceptions which are referenced by the doc blocks. 

Also I noticed we are not always checking IsGranted on the endpoints but maybe that on purpose?

src/Class/Controller/CustomLayout/UpdateController.php
```diff
@@ -0,0 +1,86 @@
...
+    )]
+    #[CustomLayoutUpdateRequestBody]
+    #[SuccessResponse(
+        description: 'pimcore_studio_api_class_custom_layout_update_success_response',
+        content: new JsonContent(ref: CustomLayout::class, type: 'object')
+    )]
+    #[DefaultResponses([
+        HttpResponseCodes::NOT_FOUND,
+    ])]
+    public function updateVersion(
```
lukmzig: should this be updateCustomLayout?

src/Class/Schema/CustomLayout/CustomLayout.php
```diff
@@ -0,0 +1,114 @@
...
+    required: [
+        'id',
+        'name',
+        'description',
+        'creationDate',
+        'modificationDate',
+        'userOwner',
+        'classId',
+        'default',
+        'blockedVarsForExport',
```
lukmzig:
```suggestion
        'layoutDefinition',
```

lukmzig (approved):
LGTM :)

PR #730, opened 2025-01-28
[Task] Change dataobject preview route to get

mattamon:


Corepex (approved):
works for me :100:  - thx

PR #733, opened 2025-01-28
[Grid] Add Asset Preview Resolver

### LINKED ISSUE #726: Adjust grid endpoint to improve `preview` column
Endpoint: `/pimcore-studio/api/assets/grid`

If the preview column is there, we return the path to the file. 
That works for images but not for files that are not image-based (like .mp3 or .pdf)

We should return a thumbnail or, if not possible, an icon there. 
Afaik we already do that in `/api/assets/tree` see `imageThumbnailPath` and `icon`

martineiber:
Resolves #726


mattamon (approved):

Corepex (approved):
LGTM - thx :100:

PR #740, opened 2025-02-03
[Grid] [DataObject] Should use Database as Source

### LINKED ISSUE #738: [Grid] [DataObject] Should use Database as Source
DataObject Grid Should use Database as Source to display the data. But only for the filtered result set. Search will still be done by the Index. Asset Grid should stay with the Index

Part of https://github.com/pimcore/studio-ui-bundle/issues/296

martineiber:
Resolves #738


lukmzig (approved):
LGTM

PR #742, opened 2025-02-04
[Data Objects] Consider custom layouts in the data object layout endpoint

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/328

### DIFF
--- a/src/Security/Service/LayoutService.php
+++ b/src/Security/Service/LayoutService.php
@@ -29,7 +29,6 @@ use function sprintf;
  */
 final readonly class LayoutService implements LayoutServiceInterface
 {
-
     public function getUserAllowedLayoutsByClass(DataObject $dataObject, UserInterface $user): array
     {
         $allowedLayouts = $this->getUserAllowedLayouts($dataObject, $user);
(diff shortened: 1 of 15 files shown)


src/Security/Service/LayoutService.php
```diff
@@ -0,0 +1,56 @@
...
+                $userPermissions[] = $match[2];
+            }
+        }
+
+        return $userPermissions;
+    }
+
+    public function getUserAllowedLayouts(DataObject $dataObject, UserInterface $user): array
+    {
+        /** @var User $user */
```
martineiber, line 48: We could add a `instance of User` check? Then we do not need the PHP Comment.
WDYT?

martineiber (changes requested):

src/Security/Service/LayoutService.php
lukmzig: We could, but we have it with multiple places like this and I would not change it now.
$user will always be instance of `User` the issue is that core and also GDI are not working with UserInterface but User model directly

If we do instanceof the we also have to handle a case which never will be true (that $user is not instance of User)

martineiber (approved):

Issue #745, opened 2025-02-04
[Widgets] REST endpoints for config


lukmzig:
Create a new tag for Perspectives
Add endpoints:

- [x] add class definition list
- [x] add context menu permissions list based on the element type
- [x] list all available widget types (extendable Symfony config node)
- [x] implement Location Aware config for tree widgets
- [x] get tree widget config
- [x] create a new tree widget
- [x] update tree widget
- [x] list all widgets (name/id, type)
- [x] delete based on name/dd

PR #752, opened 2025-02-05
[Elements] Add endpoint to list context permissions

lukmzig:
Part of https://github.com/pimcore/studio-backend-bundle/issues/745
Part of https://github.com/pimcore/studio-ui-bundle/issues/311


src/Element/Controller/GetContextPermissionsController.php
```diff
@@ -0,0 +1,78 @@
...
+
+    public function __construct(
+        SerializerInterface $serializer,
+        private readonly ElementServiceInterface $elementService
+    ) {
+        parent::__construct($serializer);
+    }
+
+    /**
+     * @throws AccessDeniedException|UserNotFoundException|NotFoundException
```
martineiber: Those Exceptions seams not to be thrown in the Element Service?

martineiber (changes requested):

src/Element/Service/ElementService.php
```diff
@@ -116,6 +121,27 @@ public function getElementSubtype(ElementParameters $parameters): Subtype
...
+     * @throws InvalidElementTypeException
+     */
+    public function getElementContextPermissions(
+        string $elementType
+    ): AssetContextPermissions|DataObjectContextPermissions|DocumentContextPermissions {
+        $permissions = match ($elementType) {
+            ElementTypes::TYPE_ASSET => new AssetContextPermissions(),
+            ElementTypes::TYPE_DATA_OBJECT => new DataObjectContextPermissions(),
+            ElementTypes::TYPE_DOCUMENT => new DocumentContextPermissions(),
+            default => throw new InvalidElementTypeException($elementType),
```
martineiber, line 134: So the Context Permission is alway the same?

It Does not depend on any user Permission? Only later on the Perspectives?

lukmzig: This is the list which is applied in the config,  notnsure how would you set up config of widget per user …
But when considering permissions they should be then considered in combination with user permissions.
These are context ( tree permissions )so they depend on element type and have nothing to do with perspective itself. Perspective will introduce other permissions as well

src/Element/Controller/GetContextPermissionsController.php
```diff
@@ -0,0 +1,76 @@
...
+
+    public function __construct(
+        SerializerInterface $serializer,
+        private readonly ElementServiceInterface $elementService
+    ) {
+        parent::__construct($serializer);
+    }
+
+    /**
+     * @throws InvalidElementTypeException|UserNotFoundException
```
martineiber:
```suggestion
     * @throws InvalidElementTypeException
```

Can also be removed, right?

martineiber (changes requested):

src/Element/Controller/GetContextPermissionsController.php
lukmzig: yeah you are right :+1: , I thought #[IsGranted(UserPermissions::ELEMENT_TYPE_PERMISSION->value)] is throwing an exception but thats just part of default responses

martineiber (approved):

PR #755, opened 2025-02-06
[Task] Implemented save as version

### LINKED ISSUE #741: Implement `save as draft`
Please implement the `save as draft` function in the API.

mattamon:
Resolves #471 

## Additional info
Implemented save as version and autoSave

Updated structure

### DIFF
--- a/config/data_index_filters.yaml
+++ b/config/data_index_filters.yaml
@@ -50,6 +50,9 @@ services:
   Pimcore\Bundle\StudioBackendBundle\DataIndex\Filter\FullTextFilter:
     tags: [ 'pimcore.studio_backend.search_index.filter' ]
 
+  Pimcore\Bundle\StudioBackendBundle\DataIndex\Filter\SortFilter:
+    tags: [ 'pimcore.studio_backend.search_index.filter' ]
+
   # DataObject
   Pimcore\Bundle\StudioBackendBundle\DataIndex\Filter\DataObject\ClassNameFilter:
       tags: [ 'pimcore.studio_backend.search_index.data_object.filter' ]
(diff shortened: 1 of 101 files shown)


lukmzig (commented):
Do we really need to pass user in Updater Adapters? we can request him from security service or? Updater is executed only on 1 element at the time directly so not using GEE?

mattamon:
> Do we really need to pass user in Updater Adapters? we can request him from security service or? Updater is executed only on 1 element at the time directly so not using GEE?

For publish we need it, at least this is why I added it. I mean it does not hurt, but of course we could always get it from the request.

lukmzig:
> > Do we really need to pass user in Updater Adapters? we can request him from security service or? Updater is executed only on 1 element at the time directly so not using GEE?
> 
> For publish we need it, at least this is why I added it. I mean it does not hurt, but of course we could always get it from the request.

Yeah, I meant in this case we do not need to pass it down all the time for every adapter since we need it only in 1 which can get it from the request.

mattamon:
> > > Do we really need to pass user in Updater Adapters? we can request him from security service or? Updater is executed only on 1 element at the time directly so not using GEE?
> > 
> > 
> > For publish we need it, at least this is why I added it. I mean it does not hurt, but of course we could always get it from the request.
> 
> Yeah, I meant in this case we do not need to pass it down all the time for every adapter since we need it only in 1 which can get it from the request.

Yep I totally get it, currently two adapters needed it, but I will revert it :)

lukmzig (approved):
LGTM :)

markus-moser:
@mattamon Thanks a lot, I gave it a try and it works like a charme - I tested both `autoSave` and `version` and it worked as expected. Cool that you already respected both cases. 👍 

Can we merge it?

PR #772, opened 2025-02-10
[Task] Open element

### LINKED ISSUE #756: Add endpoints for open asset/data object/document
Image

Should work similar like in the classic bundle:

`/admin/element/get-subtype?_dc=1738839102001&id=%2FCar%20Images%2Fjaguar%2Fauto-30951195555.jpg&type=asset`

* It should search by default via ID/full path (please check the old code if other fields are respected)
* The search/look up mechanism should be extendable via events - e.g. to make it possible to search by an article number data object attribute (we also have an event for that in the classic bundle)
* We only need the ID as response and no sub-type etc.

mattamon:
Resolves #756 

## Additional info
Adding resolve endpoint with event that can change the full path or ID


src/DataObject/Legacy/ApplyChangesHelper.php
```diff
@@ -66,6 +67,7 @@ public function applyChanges(Concrete $object, array $changes): void
 
             if ($fd) {
                 if ($fd instanceof Localizedfields) {
+                    /** @var User $user */
                     $user = $this->securityService->getCurrentUser();
                     if (!$user->getAdmin()) {
                         $allowedLanguages = $this->dataObjectServiceResolver->getLanguagePermissions(
```
lukmzig, line 73: I think we should replace this completely with our internal service:
https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Security/Service/LanguageService.php#L36

mattamon: Agreed, let's do it in a follow up PR since this is actually not part of this PR.

lukmzig (approved):

PR #778, opened 2025-02-12
[Data Object Editor] Layout endpoint should return main class layout by default

### LINKED ISSUE #775: [Data Object Editor] User needs to have access to default/main layout when no layout is defined in workspace
Currently in that case it's impossible to fetch the layout whereas in the classic bundle it's possible to display the data object with the default/main layout.

Image

### LINKED ISSUE #776: Unauthorized vs. forbidden response code
Currently e.g. in this case we send a 401 response:

https://github.com/pimcore/studio-backend-bundle/issues/775

Can we change permission error response codes to 403 and use 401 really for unauthenticated? The background is that we have a mechanism in place which redirects to the login form if we receive 401 response codes as we asume that we where logged out. 

Therefore a proper use of response codes would help at this point.

https://beeceptor.com/docs/concepts/401-vs-403/#:~:text=401%20indicates%20missing%20or%20invalid,of%20permission%20despite%20valid%20authentication.

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/775
Resolves https://github.com/pimcore/studio-backend-bundle/issues/776

### DIFF
--- a/src/DataObject/Controller/LayoutController.php
+++ b/src/DataObject/Controller/LayoutController.php
@@ -21,7 +21,6 @@ use OpenApi\Attributes\JsonContent;
 use Pimcore\Bundle\StudioBackendBundle\Controller\AbstractApiController;
 use Pimcore\Bundle\StudioBackendBundle\DataObject\Schema\Layout;
 use Pimcore\Bundle\StudioBackendBundle\DataObject\Service\LayoutServiceInterface;
-use Pimcore\Bundle\StudioBackendBundle\Exception\Api\AccessDeniedException;
 use Pimcore\Bundle\StudioBackendBundle\Exception\Api\ForbiddenException;
 use Pimcore\Bundle\StudioBackendBundle\Exception\Api\InvalidElementTypeException;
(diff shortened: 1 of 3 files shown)


markus-moser:
@lukmzig I have a question regarding #776: Do you know if we have situations like that also for other endpoints where we use 401 vs 403? Is it possible to do a quick check in the codebase here?

mattamon:
> @lukmzig I have a question regarding #776: Do you know if we have situations like that also for other endpoints where we use 401 vs 403? Is it possible to do a quick check in the codebase here?

Already talked about this, and yes we found places where we need to replace it.

lukmzig:
> @lukmzig I have a question regarding #776: Do you know if we have situations like that also for other endpoints where we use 401 vs 403? Is it possible to do a quick check in the codebase here?

https://github.com/pimcore/studio-backend-bundle/pull/780
added a separate PR as we had to change a lot ...

markus-moser:
@lukmzig I tested the custom layouts and everything works well now 👍

martineiber (approved):

PR #779, opened 2025-02-12
[Task] Add draft data

### LINKED ISSUE #771: [Data Objects] Add autoSave draft information on detail endpoint
Please add a flag (or similar) to the detail endpoint to make it possible to identify auto saved drafts.

`/pimcore-studio/api/data-objects/{ID}`

mattamon:
Resolves #771 

## Additional info
Changed method getLatestVersion to really get the version first and added another to load the concrete after.
So we do not need to load the version twice.

### DIFF
--- a/src/DataObject/Schema/DataObject.php
+++ b/src/DataObject/Schema/DataObject.php
@@ -43,6 +43,7 @@ use Pimcore\Bundle\StudioBackendBundle\Util\Trait\WorkflowAvailableTrait;
         'childrenSortBy',
         'childrenSortOrder',
         'objectData',
+        'draftData',
         'inheritanceData',
         'allowInheritance',
         'allowVariants',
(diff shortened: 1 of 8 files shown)


lukmzig (changes requested):
https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Workflow/Service/WorkflowDetailsService.php#L109

we need to adjust this method as well right?

mattamon:
> https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Workflow/Service/WorkflowDetailsService.php#L109
> 
> we need to adjust this method as well right?

https://github.com/pimcore/studio-backend-bundle/pull/779/files#diff-f469b888f4e2ef34334883e1aa70ace5730ea31a206475ab05b2c700f55e8011R108
Changed it here already :)

lukmzig (approved):
LGTM :)

PR #790, opened 2025-02-12
Versions] Delete version does not seem to work

### LINKED ISSUE #788: [Versions] Delete version does not seem to work for data object versions
I get a 403 response although I am logged in as admin - for assets it works.

Image

lukmzig:
Resolves #788


src/Version/Controller/DeleteController.php
```diff
@@ -51,7 +50,6 @@ public function __construct(
      * @throws ForbiddenException|NotFoundException|UserNotFoundException
      */
     #[Route('/versions/{id}', name: 'pimcore_studio_api_delete_version', methods: ['DELETE'])]
-    #[IsGranted(ElementPermissions::VERSIONS_PERMISSION)]
```
martineiber, line 54: So there will be no user permission check at all?

lukmzig: There is but on the element
https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Version/Controller/DeleteController.php#L77

there is no user permission related to versions

martineiber (approved):

PR #815, opened 2025-02-20
Data Object Editor] Video poster image wrong field

### LINKED ISSUE #813: [Data Object Editor] Video poster image cannot be saved
When I try to update the video poster image with the same format like in the `GET` endpoint an exception is thrown.

```JSON
"video": {
      "type": "asset",
      "poster": {
        "type": "asset",
        "id": 1186,
        "fullPath": "/PNG_transparency_demonstration_1.png"
      },
      "data": {
        "type": "asset",
        "id": 427,
        "fullPath": "/Sample Content/Videos/Volkswagen-Van_1.mp4",
        "subtype": "video"
      }
    },
```

It seems that the update currently expects a path only. Can you please change it to the same format like the video (in `data`)?

lukmzig:
Resolves #813

### DIFF
--- a/src/DataObject/Data/Adapter/VideoAdapter.php
+++ b/src/DataObject/Data/Adapter/VideoAdapter.php
@@ -69,8 +69,8 @@ final readonly class VideoAdapter implements SetterDataInterface, DataNormalizer
         $type = $adapterData['type'] ?? null;
         $poster = $adapterData['poster'] ?? null;
         if ($type === ElementTypes::TYPE_ASSET) {
-            $adapterData['data'] = $this->resolveAssetIfNeeded($type, $adapterData['data']['fullPath']);
-            $adapterData['poster'] = $this->getAssetByPath($poster ? $poster['fullPath'] : null);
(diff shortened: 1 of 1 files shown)


martineiber (approved):

src/DataObject/Data/Adapter/VideoAdapter.php
```diff
@@ -67,9 +67,10 @@ public function getDataForSetter(
         }
 
         $type = $adapterData['type'] ?? null;
+        $poster = $adapterData['poster'] ?? null;
         if ($type === ElementTypes::TYPE_ASSET) {
             $adapterData['data'] = $this->resolveAssetIfNeeded($type, $adapterData['data']['fullPath']);
-            $adapterData['poster'] = $this->getAssetByPath($adapterData['poster'] ?? null);
+            $adapterData['poster'] = $this->getAssetByPath($poster ? $poster['fullPath'] : null);
```
markus-moser: Not a massive problem, but shouldn't we resolve the asset relation for `poster` and `data` by ID instead of the path? If the asset gets moved in the meantime the ID would still work but the path would be broken...

PR #833, opened 2025-02-25
[Perspectives] Add default widget icon and add PQL filter in asset tree

### LINKED ISSUE #830: [Asset Tree] Enable PQL also for assets
We now have a `pqlQuery` filter in the data objects tree endpoint. We would need the same also for the assets tree to support asset tree widgets with PQL filter.

### LINKED ISSUE #832: [Perspectives] Default perspective icons
Currently the default perspectives are set to `null` in the default perspectives config yaml. This somehow leads to a fallback icon like this:

```json
{
   "type": "name",
   "icon": "unknown"
}
```

A fallback to the `unknown` icon if the icon is undefined is in general a good idea. Nevertheless for the default perspectives we should define the right icons which are as following:

`asset`, `document` or `data-object`(with type `name`)

Unfortunately it's not easily possible to this fallback logic on frontend side (so that we map it based on the element type of the tree). But I'd say a fallback to `unknown` instead of a element type specific fallback is fine.

Therefore please change the icons of the default perspective.

lukmzig:
Resolves #832
Resolves #830

## Additional info
Add pql filter to asset tree
Add default tree widget icons

### DIFF
--- a/src/Asset/Controller/TreeController.php
+++ b/src/Asset/Controller/TreeController.php
@@ -83,7 +83,7 @@ final class TreeController extends AbstractApiController
     #[TextFieldParameter(
         name: 'pqlQuery',
         description: 'Pql query filter',
-        example: 'id = 150'
+        example: 'id = 69'
     )]
     #[ExcludeFoldersParameter]
     #[PathParameter]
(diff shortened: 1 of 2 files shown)


src/Asset/Controller/TreeController.php
```diff
@@ -79,6 +80,11 @@ public function __construct(
         example: 1,
     )]
     #[IdSearchTermParameter]
+    #[TextFieldParameter(
+        name: 'pqlQuery',
+        description: 'Pql query filter',
+        example: 'id = 150'
```
mattamon:
```suggestion
        example: 'id = 69'
```

mattamon (changes requested):

martineiber (approved):

mattamon (approved):
😼

PR #837, opened 2025-02-25
[Perspectives] Update perspective configuration

### LINKED ISSUE #834: [Perspectives] Problem with schema of PerspectiveConfigDetail
`PerspectiveConfigDetail` has a `isWriteable` property with a different type compared to the `PerspectiveConfig`. Therefore the API generation produces an error becaue the types don't match.

Image

I guess we just can remove this poperty from the `PerspectiveConfigDetail` schema?

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/308
Resolves #834

## Additional info
Add endpoint and validations for perspective update
Add small fix for perspective config schema

### DIFF
--- a/src/Perspective/Controller/UpdateController.php
+++ b/src/Perspective/Controller/UpdateController.php
@@ -59,7 +59,7 @@ final class UpdateController extends AbstractApiController
         name: 'pimcore_studio_api_update_perspectives_config',
         methods: ['PUT']
     )]
-    #[IsGranted(UserPermissions::WIDGET_EDIT->value)]
+    #[IsGranted(UserPermissions::PERSPECTIVE_EDITOR->value)]
     #[Put(
         path: self::PREFIX . self::ROUTE,
         operationId: 'perspective_update_config_by_id',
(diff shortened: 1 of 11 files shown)


src/Perspective/Controller/UpdateController.php
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * @throws ElementSavingFailedException|InvalidArgumentException|NotFoundException|NotWriteableException
+     */
+    #[Route(
+        self::ROUTE,
+        name: 'pimcore_studio_api_update_perspectives_config',
+        methods: ['PUT']
+    )]
+    #[IsGranted(UserPermissions::WIDGET_EDIT->value)]
```
martineiber: Should be PERSPECTIVE_EDITOR? 
```suggestion
    #[IsGranted(UserPermissions::PERSPECTIVE_EDITOR->value)]
```

martineiber (changes requested):

src/Perspective/Controller/UpdateController.php
lukmzig: you are right, thanks :)

martineiber (approved):

PR #838, opened 2025-02-25
[Tree Widget] Add rootFolderId

### LINKED ISSUE #835: [Tree Widget] Add rootFolderId if possible
Currently we get the `rootFolder` config as path as it is configured by path in our config. That's fine! But is it possible to send in our perspectives endpoint also the ID of the folder in addition?

Our tree works with (parent) IDs in the frontend. In the tree endpoints we would also have the paths but it's not easy to refactor it in the frontend to a path logic. Therefore at the moment I just use our `get id by path` endpoint to fetch the ID.

If it is too complex to add this it's fine with the additional call but we would save one request if you add it directly. In general it could be a benefit if a widget has the chance to "enrich" the sent data in addition to the config - but we should not invest too much time for now if this is tricky.

lukmzig:
Resolves #835

## Additional info
- Add root folder ID to the tree widget response
- Validate root path on tree widget update

### DIFF
--- a/src/Perspective/Hydrator/ElementTreeWidgetConfigHydrator.php
+++ b/src/Perspective/Hydrator/ElementTreeWidgetConfigHydrator.php
@@ -16,10 +16,15 @@ declare(strict_types=1);
 
 namespace Pimcore\Bundle\StudioBackendBundle\Perspective\Hydrator;
 
+use Pimcore\Bundle\StudioBackendBundle\Element\Request\PathParameter;
+use Pimcore\Bundle\StudioBackendBundle\Element\Service\ElementServiceInterface;
 use Pimcore\Bundle\StudioBackendBundle\Element\Service\Permissions\ContextPermissionServiceInterface;
+use Pimcore\Bundle\StudioBackendBundle\Exception\Api\NotFoundException;
 use Pimcore\Bundle\StudioBackendBundle\Icon\Service\IconServiceInterface;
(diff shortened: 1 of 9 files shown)


src/Perspective/Hydrator/ElementTreeWidgetConfigHydrator.php
```diff
@@ -49,11 +57,32 @@ public function hydrate(array $widgetData): ElementTreeWidgetConfig
...
             $widgetData['classes'],
             $widgetData['pql'],
             $widgetData['pageSize'],
             $widgetData['isWriteable']
         );
     }
+
+    private function getRootFolderId(array $widgetData): int
+    {
+        $folderId = 1;
```
martineiber: The default folder is defined twice with 1. 
Here as well https://github.com/pimcore/studio-backend-bundle/pull/838/files#diff-6b126f2f236354af50b05b0b6a448af0dd7c3c4ff0808cb1eac020673e8bbfd6R67

Do we have somehow the possibility to reuse it?

martineiber (approved):

src/Perspective/Hydrator/ElementTreeWidgetConfigHydrator.php
lukmzig: You mean with some kind of ENUM? Otherwise I do not understand the problem

martineiber: Yeah ENUM would make sense.

PR #848, opened 2025-02-26
[Data Object Editor] ValidationExceptions

### LINKED ISSUE #819: [Data Object Editor] Status code of ValidationExceptions
Pimcore throws for validation errors in a lot of cases a `ValidationException`, e.g. here:

https://github.com/pimcore/pimcore/blob/8ce5e6d9b83363a7e419de99ae4ed19f69a7ed6e/models/DataObject/ClassDefinition/Data/Input.php#L218

This is mainly relevant for the data object editor to validate mandatory fields and other types of validations. In the data object update endpoint this exception is handled the right way, so we get the error message etc. But could you please catch the 500 status code and send a different suitable one in the 400 range (e.g. 422)?

Don't know if it possible to handle this on your side on a very central place so that `ValidationException`s are handled always with the new status code. If yes, this would be a very good benefit - otherwise it should be respected at least in the update endpoint.

lukmzig:
Resolves #819 

## Additional info
- add new validation exception for editable fields
- add new translation key

### DIFF
--- a/src/Element/Service/ElementSaveService.php
+++ b/src/Element/Service/ElementSaveService.php
@@ -24,7 +24,6 @@ use Pimcore\Bundle\StudioBackendBundle\Util\Constant\ElementPermissions;
 use Pimcore\Bundle\StudioBackendBundle\Util\Constant\ElementSaveTasks;
 use Pimcore\Model\DataObject\Concrete;
 use Pimcore\Model\Document;
-use Pimcore\Model\Element\DuplicateFullPathException;
 use Pimcore\Model\Element\ElementInterface;
 use Pimcore\Model\UserInterface;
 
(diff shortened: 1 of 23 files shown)


src/Element/Service/ElementSaveService.php
```diff
@@ -40,8 +39,7 @@ public function __construct(
     }
 
     /**
-     * @throws DuplicateFullPathException
-     * @throws Exception
+     * @inheritdoc
```
martineiber:
```suggestion
     * {@inheritdoc}
```
TBH i have no idea what's the difference, but docs says it should be with {}
https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.inlineinheritdoc.pkg.html

martineiber (approved):

src/Element/Service/ElementSaveService.php
lukmzig: Hmm I did not see, that, at some places, e.g here we also have it like `inheritDoc`
https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Element/Service/ElementLocationService.php#L52

I will replace them to the format you posted

martineiber: At least we should keep it consistent...

lukmzig: now they should be consistent :)

PR #851, opened 2025-02-27
[Tree Widget] Use translation keys for tree widgets of default perspective

markus-moser:
As those labels should be localized, we use the name as translation key...

### DIFF
--- a/config/pimcore/default_perspective.yaml
+++ b/config/pimcore/default_perspective.yaml
@@ -20,7 +20,7 @@ studio_default_perspective:
       isWriteable: false
     studio_asset_tree_widget:
       id: "studio_asset_tree_widget"
-      name: "Asset Tree"
+      name: "asset.asset-tree.title"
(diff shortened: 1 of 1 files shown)


config/pimcore/default_perspective.yaml
```diff
@@ -5,7 +5,7 @@ studio_default_perspective:
   widgetsLeft:
     studio_document_tree_widget:
       id: "studio_document_tree_widget"
-      name: "Document Tree"
+      name: "asset.asset-tree.title"
```
lukmzig:
```suggestion
      name: "document.document-tree.title"
```

lukmzig (changes requested):
I think asset tree name is still missing, got mixed up with document tree

markus-moser:
@lukmzig Thanks a lot for checking, now it should be fine :)

lukmzig (approved):

PR #904, opened 2025-03-17
[Task] Use save service

### LINKED ISSUE #895: [DataObject] Creation shouldnt be async
Hi Guys, 

I just used the `/api/data-objects/add/{parentId}` endpoint and noticed that the response brings back the ID of the created object, but it's not there when clearing the cache (which recalls the tree endpoint). 

We had that issue in multiple places ... 

Could you please change the behaviour so that the response returns the ID if the object is already created (after the queue)?

Hint: Last time, we had to enable the `enablesynprocessing` in GDI

mattamon:
Resolves #895 

## Additional info
Enable sync processing for adding new objects


martineiber (approved):
LGTM

PR #911, opened 2025-03-18
[Search] [Data Objects] Get Default grid Columns for search

### LINKED ISSUE #885: [Search] [Data Objects] Get Default grid Columns for search
Part of https://github.com/pimcore/studio-ui-bundle/issues/247

see https://github.com/pimcore/studio-backend-bundle/issues/691

martineiber:
Resolves #885

### DIFF
--- a/src/DataObject/Controller/Search/GetConfigurationController.php
+++ b/src/DataObject/Controller/Search/GetConfigurationController.php
@@ -77,7 +77,7 @@ final class GetConfigurationController extends AbstractApiController
         HttpResponseCodes::NOT_FOUND,
         HttpResponseCodes::UNAUTHORIZED,
     ])]
-    public function getDataObjectGridConfiguration(?string $classId): JsonResponse
+    public function getDataObjectSearchConfiguration(?string $classId): JsonResponse
     {
         $configuration = $this->gridConfigurationService->getDataObjectSearchConfiguration($classId);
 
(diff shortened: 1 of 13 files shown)


src/DataObject/Controller/Search/GetConfigurationController.php
```diff
@@ -0,0 +1,86 @@
...
+        private readonly ConfigurationServiceInterface $gridConfigurationService,
+    ) {
+        parent::__construct($serializer);
+    }
+
+    /**
+     * @throws NotFoundException
+     */
+    #[Route(
+        '/data-object/search/configuration/{classId?}',
```
lukmzig, line 52:
```suggestion
        '/data-object/search/configuration/{classId}',
```
should it be like this?

martineiber: Nop.. classId has to be optional, that's why there is a `?`

lukmzig: Ah I see ... but the does it not need to be the same here?
https://github.com/pimcore/studio-backend-bundle/pull/911/files#diff-91c31968b6e99f9f0907012a0e6020afd877a5d981f7563472140d4a14b9e837R59

martineiber: This does not work, Swagger can not work with `?`
It is defined here, that it is optional
https://github.com/pimcore/studio-backend-bundle/pull/911/files#diff-91c31968b6e99f9f0907012a0e6020afd877a5d981f7563472140d4a14b9e837R69

lukmzig: Alright, thanks for the clarification! :)

lukmzig (approved):
LGTM :)

PR #913, opened 2025-03-18
Added allowed classes and allowed doctypes

### LINKED ISSUE #910: [Class] Permissions not prespected in classes endpoint
The endpoint `/api/class/collection` always returns all classes.

### How to reproduce
1. create a user
2. give object permissions
3. select only one class under "classes"
4. give "view" and "list" workspace permissions on object root

Now the endpoint should only return the one class you have selected but it returns all.

mcop1:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/910

## Additional info
Added doctypes as well, will be needed in the future.


Corepex (approved):
LGTM

PR #918, opened 2025-03-19
[Search] [Data Objects] Search Endpoint

### LINKED ISSUE #892: [Search] [Data Objects] Search Endpoint
Similar to `/pimcore-studio/api/data-objects/grid/{classId}`

- Class Id can be optional
- Include Folders
- Filter for types
    - Object
    - Folder
    - Variant 

Part of https://github.com/pimcore/studio-ui-bundle/issues/247

martineiber:
Resolves #892

### DIFF
--- a/config/data_index_filters.yaml
+++ b/config/data_index_filters.yaml
@@ -53,6 +53,9 @@ services:
   Pimcore\Bundle\StudioBackendBundle\DataIndex\Filter\SortFilter:
     tags: [ 'pimcore.studio_backend.search_index.filter' ]
 
+  Pimcore\Bundle\StudioBackendBundle\DataIndex\Filter\StringFilter:
+    tags: [ 'pimcore.studio_backend.search_index.filter' ]
+
   # DataObject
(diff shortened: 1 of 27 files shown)


lukmzig (approved):
LGTM

Only thing I am not competely sure is the TAG itself, based on the original issue:
https://github.com/pimcore/studio-ui-bundle/issues/245

I created a search tag (namespace), the idea was that this one will have all searches, currently it contains the generic search and search previews based on the element types.  Not sure if specific search should be there based on the element type as well or rather under specific element namespace.

martineiber:
> LGTM
> 
> Only thing I am not competely sure is the TAG itself, based on the original issue: [pimcore/studio-ui-bundle#245](https://github.com/pimcore/studio-ui-bundle/issues/245)
> 
> I created a search tag (namespace), the idea was that this one will have all searches, currently it contains the generic search and search previews based on the element types. Not sure if specific search should be there based on the element type as well or rather under specific element namespace.

Good Point! Did not see that.
I will move all new endpoints to that tag

lukmzig (approved):

PR #927, opened 2025-03-20
[Perspectives] Integrate into user management

### LINKED ISSUE #144: [User Management] [Endpoint] Available User Perspectives
Part of https://github.com/pimcore/studio-backend-bundle/issues/138

### LINKED ISSUE #836: [Perspectives] Integrate into user management
The information which perspectives are available for a user and which perspective is the currently active one is stored in the user table for the perspectives in the classic bundle (`perspectives` and `activePerspective` columns).

We need to find a way to integrate the logic for the new perspectives too. Please keep in mind that it is needed to somehow make it possible that studio UI and classic bundle work in parallel. I'd say it's ok when setting the perspectives permissions is only possible in the related UI (so for classic bundle in classic bundle and for studio in studio). Not sure if extending the user entity in the core for studio is a good idea or if it would be better to create a new additional table within studio? With this approach we could keep the core clean from studio related things (and at some point after sunsetting the classic bundle we could deprecate the properties in the core...).

lukmzig:
Resolves #144 
Resolves #836 

## Additional info
- Add separate table to store studio perspectives user data
- Added info about available and active perspectives to user info endpoint
- Added possibility to read and update studio perspectives with user/role management
- Added endpoint to update user active perspective
- Adapted user permission checks based on the studio perspectives

### DIFF
--- a/src/User/Service/UserUpdateService.php
+++ b/src/User/Service/UserUpdateService.php
@@ -23,7 +23,9 @@ use Pimcore\Bundle\StudioBackendBundle\Exception\Api\DatabaseException;
 use Pimcore\Bundle\StudioBackendBundle\Exception\Api\ForbiddenException;
 use Pimcore\Bundle\StudioBackendBundle\Exception\Api\InvalidArgumentException;
 use Pimcore\Bundle\StudioBackendBundle\Exception\Api\NotFoundException;
+use Pimcore\Bundle\StudioBackendBundle\Exception\Api\NotWriteableException;
 use Pimcore\Bundle\StudioBackendBundle\Exception\Api\ParseException;
+use Pimcore\Bundle\StudioBackendBundle\Exception\Api\UserNotFoundException;
 use Pimcore\Bundle\StudioBackendBundle\Security\Service\SecurityServiceInterface;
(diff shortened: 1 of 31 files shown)


src/User/Service/UserUpdateService.php
```diff
@@ -128,6 +132,15 @@ public function updatePasswordById(UpdatePasswordParameter $updateParameter, int
         $this->userRepository->updateUser($user);
     }
 
+    /**
+     * @throws ForbiddenException|NotFoundException|NotWriteableException|UserNotFoundException
```
mattamon, line 136: Should we put all that throw statements in the interface and then just use inherit?

lukmzig: not sure why? 
Its adding 2 more exception classes but there are more used in the class:
https://github.com/pimcore/studio-backend-bundle/blob/a9a181e2782d8db3c472e3dd5dd7c21e8d743c4a/src/User/Service/UserUpdateService.php

I would say it does not make sense to do it for 1 method IMHO, but either make it consistent for all methods and all classes or when its necessary (e.g some big classes ...)

martineiber (approved):
LGTM

PR #937, opened 2025-03-24
[Task] Creation date can be null

mattamon:
See title


lukmzig (changes requested):
can we change also $modificationDate to nullable?

lukmzig (approved):

PR #947, opened 2025-03-25
[Translations] Add missing translation endpoints

### LINKED ISSUE #930: [Translations] New endpoint for missing translations
Hi guys, 

We need a new endpoint that handles missing translations. 
Like in the old UI it should add the missing translations to the translations panel.

IMHO, it doesn't make sense to call the endpoint for each single translation key .... it would be nice to send an array of keys each time.

lukmzig:
Resolves #930

## Additional info
- add endpoint to add/update studio  translations
- add endpoint to delete studio translations
- update list endpoint to return also data form the database

### DIFF
--- a/src/Translation/Schema/UpdateTranslation.php
+++ b/src/Translation/Schema/UpdateTranslation.php
@@ -19,7 +19,6 @@ namespace Pimcore\Bundle\StudioBackendBundle\Translation\Schema;
 use OpenApi\Attributes\Items;
 use OpenApi\Attributes\Property;
 use OpenApi\Attributes\Schema;
-use Pimcore\Bundle\StudioBackendBundle\Util\Constant\PublicTranslations;
 
 /**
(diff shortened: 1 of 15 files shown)


src/Translation/Schema/UpdateTranslation.php
```diff
@@ -0,0 +1,55 @@
...
+use OpenApi\Attributes\Schema;
+
+/**
+ * @internal
+ */
+#[Schema(
+    schema: 'UpdateTranslation',
+    title: 'Translation Update',
+    description: 'Translation Update Scheme for API',
+    required: ['locale', 'keys'],
```
martineiber:
```suggestion
    required: ['locale', 'translationData'],
```

martineiber (changes requested):

mattamon (approved):
LGTM

Issue #950, opened 2025-03-26
[Asset Metadata] Implement adapters


lukmzig:
The adapter approach is introduced for the asset metadata field types

Have a look if all core adapters are processing values correctly, if not add studio adapter (e.g for relations)

PR #951, opened 2025-03-26
[Asset Metadata] Implement adapter approach

### LINKED ISSUE #941: [Asset Metadata] Implement adapter approach
Currently when calling `/pimcore-studio/api/assets/{ID}/custom-metadata` for an object with user data type from the asset metadata classdefinitions bundle the following exeception gets triggered:

Image

The reason is because the user is passed as user object to the serializer and therefore it cannot be serialized.

After checking it a bit in the code I found out that the classic bundle uses the following method to extract the metadata:

`Asset\Service::expandMetadataForEditmode($asset->getMetadata())`

We need a similar mechanism also for studio so that it's possible to extend the metadata types.

lukmzig:
Resolves #941 

## Additional info

- User adapter approach for metada data processing (normalize and denormalize)

1. first load studio adapters
2. if studio adapter does not exists load core adapter
3. if that is not existing use raw value

- Use adapters in hydrators and updater and patcher

### DIFF
--- a/config/pimcore/config.yaml
+++ b/config/pimcore/config.yaml
@@ -99,6 +99,8 @@ pimcore_studio_backend:
             document: ['content', 'seo', 'warning', 'notice']
             data-object: ['content', 'seo', 'warning', 'notice']
 
+    asset_metadata_adapter_mapping: []
+
     data_object_data_adapter_mapping:
         Pimcore\Bundle\StudioBackendBundle\DataObject\Data\Adapter\AdvancedManyToManyRelationAdapter:
             - "advancedManyToManyRelation"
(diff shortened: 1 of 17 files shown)


config/pimcore/config.yaml
```diff
@@ -99,6 +99,8 @@ pimcore_studio_backend:
             document: ['content', 'seo', 'warning', 'notice']
             data-object: ['content', 'seo', 'warning', 'notice']
 
+    asset_metadata_adapter_mapping: []
```
martineiber, line 102: Can we maybe a short doc how to set the mapping and how to use the `pimcore.studio_backend.metadata_adapter` Tag?

martineiber (changes requested):

config/pimcore/config.yaml
lukmzig: We will need to add some adapters, eg. for relations anyway (see https://github.com/pimcore/studio-backend-bundle/issues/950)
I want to add docs as part of that issue

martineiber (approved):
LGTM

PR #958, opened 2025-03-28
Add default dummy value for JWT to ease installation

fashxp:


src/DependencyInjection/Configuration.php
```diff
@@ -288,6 +288,7 @@ private function addMercureConfiguration(ArrayNodeDefinition $node): void
                     ->scalarNode('jwt_key')
                         ->info('The key used to sign the JWT token. Must be longer than 256 bits.')
                         ->isRequired()
+                        ->defaultValue('some-secret-default')
```
jdreesen, line 291: I don't think this is a good idea. People will forget to change this and will therefore be vulnerable.

fashxp: Got a better idea to make installation run through? Right now you get an error message that container cannot build since a parameter is missing. 

So, I see two options: 
a) Either force developer to deal with mercure config before installer is called or
b) use that default value so that installation runs through and mercure can be configured properly afterwards

Mercure won't work with that key anyway since it is too short, so they need to change it anyway.

jdreesen: I see. Then it's probably okay.

fashxp: But I'm with you, doesn't look pretty. So I'm open for better suggestions :smile:

Issue #923, opened 2025-03-19
[Data Object Editor] Endpoints for classification store


markus-moser:
We need the required endpoints to be able to show the classification store fields/groups in the data object editor and to render the modal:

![Image](https://github.com/user-attachments/assets/3b69ff70-35a9-43f0-a6c4-5211862859b5)

We should analyze the endpoints which we have in the classic bundle to see what we need exactly.

martineiber:
- `admin/classificationstore/collections?storeId=1&oid=408&fieldname=technicalAttributes&page=1&start=0&limit=25`

 - `admin/classificationstore/groups?storeId=1&oid=408&fieldname=technicalAttributes&page=1&start=0&limit=25`
 - `admin/classificationstore/search-relations?storeId=1&oid=408&fieldname=technicalAttributes&page=1&start=0&limit=25`

PR #984, opened 2025-04-02
[Task] Add classification store group key Relation

martineiber:
Part of #923


lukmzig (changes requested):
I think we are missing translations for the new controller right?

martineiber:
> I think we are missing translations for the new controller right?

Good Catch, thx! I have added the translation
[e0f5c8e](https://github.com/pimcore/studio-backend-bundle/pull/984/commits/e0f5c8ea1ce5885f943bc5f3d1dd097cff16e0aa)

lukmzig (approved):
LGTM

PR #986, opened 2025-04-02
[Asset Metadata] Add metadata type to patcher schema

### LINKED ISSUE #983: [Asset Metadata] Add metadata type to patcher schema
When I send a payload like this to the `/pimcore-studio/api/assets` `PATCH` endpoint I get an exception. 

```json
[
  {
    "id": 3,
    "metadata": [
      {
        "name": "license",
        "language": null,
        "data": "test"
      }
    ]
  }
]
```

This happens because the type of the metadata field is required. But the problem is that the type is not allowed in the schema. When I overrule the schema and send the type everything is fine, so please add it to the schema.

lukmzig:
Resolves #983


martineiber (approved):
LGTM

PR #1021, opened 2025-04-15
Wrong isLocked flag when folder is locked

### LINKED ISSUE #823: Wrong `isLocked` flag when folder is locked
I noticed that the API returns the wrong `isLocked` state in the `/api/assets/tree` endpoint. 

**Steps to reproduce**
- Follow the description in https://github.com/pimcore/studio-ui-bundle/issues/962 to `lock and propagate` any folder.
- Request all items in the locked folder from `/api/assets/tree`
- check the `isLocked` flag

lukmzig:
Resolves #823 

## Additional info
Use service from GDI to retrieve the information if element should be considered as locked:
https://github.com/pimcore/generic-data-index-bundle/blob/2.x/src/SearchIndexAdapter/DefaultSearch/ElementLockService.php#L97

The service is trying to reproduce this behavior while working with index data: https://github.com/pimcore/pimcore/blob/12.x/models/DataObject/AbstractObject/Dao.php#L427


martineiber (approved):
LGTM

PR #1032, opened 2025-04-18
Fix hotspot image adapter with data object folder relation

markus-moser:
`isPublished()` does not exist for folders


lukmzig (approved):
LGTM

PR #1057, opened 2025-05-06
Mercure Client URL needs to have a dynamic url version

### LINKED ISSUE #1047: [Bug] Mercure Client URL needs to have a dynamic url version
### Studio API version

0.4.30

### Steps to reproduce

We need an option to configure the mercure client url dynamically as the domain of pimcore backend might be changing... 

see also same mechanism for direct edit https://docs.pimcore.com/platform/Direct_Edit/Installation/Mercure_Setup#configure-mercure-urls-optional

### Actual Behavior

in demo.pimcore.com mercure does not work ... as we don't know the client url during build time... 

### Expected Behavior

...

### LINKED ISSUE #1479: [Bug] Mercure Client URL needs to have a dynamic url version
### Studio API version

0.4.30

### Steps to reproduce

We need an option to configure the mercure client url dynamically as the domain of pimcore backend might be changing... 

see also same mechanism for direct edit https://docs.pimcore.com/platform/Direct_Edit/Installation/Mercure_Setup#configure-mercure-urls-optional

### Actual Behavior

in demo.pimcore.com mercure does not work ... as we don't know the client url during build time... 

### Expected Behavior

...

lukmzig:
Resolves pimcore/studio-ui-bundle#1479

### DIFF
--- a/src/DependencyInjection/PimcoreStudioBackendExtension.php
+++ b/src/DependencyInjection/PimcoreStudioBackendExtension.php
@@ -36,6 +36,7 @@ use Pimcore\Bundle\StudioBackendBundle\Perspective\Repository\PerspectiveConfigR
 use Pimcore\Bundle\StudioBackendBundle\Perspective\Service\WidgetServiceInterface;
 use Pimcore\Bundle\StudioBackendBundle\Perspective\Service\WidgetValidationServiceInterface;
 use Pimcore\Bundle\StudioBackendBundle\User\Service\KeyBindingServiceInterface;
+use Pimcore\Tool;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
@@ -196,6 +197,11 @@ class PimcoreStudioBackendExtension extends Extension implements PrependExtensio
             if ($container->hasParameter('pimcore_studio_backend.mercure_settings.' . $key)) {
                 continue;
             }
+
+            if ($key === 'hub_url_client' && $containerConfig['mercure_settings'][$key] === null) {
+                $containerConfig['mercure_settings'][$key] = Tool::getHostUrl() . '/hub';
+            }
+
             $container->setParameter(
                 'pimcore_studio_backend.mercure_settings.' . $key,
                 $containerConfig['mercure_settings'][$key]
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -279,8 +279,11 @@ class Configuration implements ConfigurationInterface
                         ->defaultValue('http://mercure/.well-known/mercure')
                     ->end()
                     ->scalarNode('hub_url_client')
-                        ->info('The url to the mercure hub for the (frontend) client.')
-                        ->defaultValue('http://localhost/hub')
+                        ->info(
+                            'The url to the mercure hub for the (frontend) client.'.
+                            ' If not set, default will be set to "http(s)://<PIMCORE_HOST>/hub".'
+                        )
+                        ->defaultNull()
                     ->end()
                     ->scalarNode('jwt_key')
                         ->info('The key used to sign the JWT token. Must be longer than 256 bits.')


src/DependencyInjection/PimcoreStudioBackendExtension.php
```diff
@@ -196,6 +197,11 @@ public function prepend(ContainerBuilder $container): void
             if ($container->hasParameter('pimcore_studio_backend.mercure_settings.' . $key)) {
                 continue;
             }
+
+            if ($key === 'hub_url_client' && $containerConfig['mercure_settings'][$key] === null) {
+                $containerConfig['mercure_settings'][$key] = Tool::getHostUrl() . '/hub';
```
martineiber, line 202: We should use the static resolver here.
https://github.com/pimcore/static-resolver-bundle/blob/3.x/src/Contract/Lib/ToolResolverContract.php#L80

martineiber (changes requested):

src/DependencyInjection/PimcoreStudioBackendExtension.php
lukmzig: why? :) not sure if we can ...

martineiber: Ohhh yeah this is in the DI.. sorry did not see that..

martineiber (approved):

src/DependencyInjection/PimcoreStudioBackendExtension.php
fashxp: @lukmzig @martineiber not sure we have the host name during container build time, do we? 
I think, that cannot work as expected. 

See also implementation in direct edit, where we have a service that resolves that during runtime: https://github.com/pimcore/direct-edit/blob/2.x/src/Service/MercureUrlService.php


We also should implement the same placeholder mechanism as in direct edit, see https://github.com/pimcore/direct-edit/blob/2.x/src/DependencyInjection/Configuration.php#L36: 

this is necessary when the system setup needs a mercure url different than `/hub`

lukmzig: @fashxp good point ... however, we do not use custom implementation in studio like in the direct edit but rather symfony funcitonality...therefore we cannot just resolve it at that point ... see here: https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Mercure/Service/HubService.php#L33

We will have to find an alternate approach for this

fashxp: this also affects this: https://github.com/pimcore/studio-ui-bundle/pull/520 - maybe it makes sense to include @markus-moser into the discussion as well.

fashxp: or is it just missing documentation, as on client side it seems that current domain is already used when no mercure url is set? 
https://github.com/pimcore/studio-ui-bundle/blob/0e59ae693e9c14f83b98b2541ec360e7317a0282/assets/js/src/core/app/config/app-config.ts#L39

but that won't work either, because then `/hub` is missing...

markus-moser: We could also just at the /hub to the fallback mechanism on studio UI side - maybe the easiest solution for now.

fashxp: but then `/hub` would be the mandatory URL ... but still, would be better as current situation.

lukmzig: @fashxp @markus-moser 
Can you please provide me some pointers how to proceed here? 
If we need to change just the client side, we could relatively easy change there current implementation and use custom URL resolver (for placeholder and dynamic host). Would that be enough?

If we need to also adapt the server URL dynamically than we need to re-build everything manually

Or do you have any other suggestions?

fashxp: I don't know the implementation details enough to give you a detailed answer. The requirements we need ordered by their priority: 

1) We need an option to configure Protocol & Host for client-side mercure URL dynamically (as it might be, that Pimcore backend is accessible via multiple host names)
2) Configure path of mercure - as their might be (infrastructure) setups where `/hub` cannot be used. 

I don't know what the best options are to archive at least 1), but ideally 1) and 2) ... in direct edit, both options are possible.

PR #1070, opened 2025-05-09
[Mercure]: Use custom URL service for Mercure

### LINKED ISSUE #1069: [Mercure]: make the sameSite setting changable
As platform.sh needs a different setting for the auth cookie we should make the samesite setting in the HubService changeable via config.
I would suggest default should still be strict.

lukmzig:
Part of pimcore/studio-ui-bundle#1479
Resolves pimcore/studio-backend-bundle#1069

## Additional info
- add possibility to make url host dynamic
- add possibility to make SameSite cookie attribute configurable


src/Mercure/Service/UrlServiceInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (https://www.pimcore.com)
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Mercure\Service;
+
+interface UrlServiceInterface
```
martineiber, line 16: on purpose not Internal? When yes, then we should add a short Docs how and for what to use it.

martineiber (changes requested):

src/Mercure/Service/UrlServiceInterface.php
lukmzig: Good point, :) I updated the docs here: https://github.com/pimcore/studio-backend-bundle/pull/1070/commits/d20b5b00f83b95a481e28a55c87e260f1ddbc4ad

martineiber (approved):

PR #1071, opened 2025-05-09
[Translations] Add fallback logic

### LINKED ISSUE #1011: [Translations] Add fallback logic
Hi Team, 

i just tried out to request the translations with language: `de`. 
There i noticed that all the translation strings are empty. (because no german translations are set)

IMHO it would make sense to add some sort of fallback logic on the BE side. This would save us several requests on the FE. 

So if there is no translation for a certain key also check the fallback language ...

WDYT?

martineiber:
Resolves #1011

## Additional info
Refactor to use the build in lazyInitialize() to fetch translations from the DB.


lukmzig (approved):
LGTM

PR #1088, opened 2025-05-19
[Docs] Fixed mercure configuration docs

fashxp:
followup to https://github.com/pimcore/studio-backend-bundle/pull/1070

@lukmzig I updated the docs to the behavior I noticed during my tests. Could you please double check. Also I'm not sure, if it makes sense, if the default `hub_url_server` ends with `/hub` ... of if it would be better to have a different default value than `hub_url_client` and end it with `/.well-known/mercure` ... as `/hub` probably will never fit? WDYT? 

Thx for checking.

### DIFF
--- a/doc/00_Installation.md
+++ b/doc/00_Installation.md
@@ -113,13 +113,13 @@ pimcore_studio_backend:
         # Optional configuration
 
         # The url to the mercure hub for the (frontend) client.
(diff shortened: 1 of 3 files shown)


fashxp:
the rest seems to work as expected!

lukmzig (approved):

Issue #1078, opened 2025-05-14
[Documents][Tree] Context menu


lukmzig:
Check that all options are covered:

- add folder
- add document types
- rename
- delete
- replace content
- copy & paste all options
- locking
- search & move

lukmzig:
Endpoints for context actions:

- add folder - `/pimcore-studio/api/elements/{elementType}/folder/{parentId}`
- add specific document type - `/pimcore-studio/api/documents/add/{parentId}`
  - to list available document types - `/pimcore-studio/api/documents/doc-types`
- rename - will be done via update endpoint
- delete - `/pimcore-studio/api/elements/{elementType}/delete/{id}`
- replace content - `/pimcore-studio/api/documents/{sourceId}/replace/{targetId}`
- copy & paste all options - `/pimcore-studio/api/documents/{id}/clone/{parentId}`
- locking - will be done via update endpoint
- search & move -  I assume this can be done via search endpoint and update of the parentId
- set as site - `/pimcore-studio/api/documents/site/{id}` (POST)
- remove site - `/pimcore-studio/api/documents/site/{id}` (DELETE)
- update site - `/pimcore-studio/api/documents/site/{id}` (POST)
- convert document - `/pimcore-studio/api/documents/{id}/convert/{type}`

PR #1091, opened 2025-05-20
[Documents][Tree] Context menu add/update and remove site

lukmzig:
Part of https://github.com/pimcore/studio-backend-bundle/issues/1078

## Additional info
- added endpoint to add/update site based on document ID
- added endpoint to remove site based on document ID

### DIFF
--- a/src/Document/Controller/Site/DeleteController.php
+++ b/src/Document/Controller/Site/DeleteController.php
@@ -57,7 +57,7 @@ final class DeleteController extends AbstractApiController
         summary: 'document_remove_site_summary',
         tags: [Tags::Documents->value]
     )]
-    #[SuccessResponse(description: 'document_remove_site_success_response',)]
+    #[SuccessResponse(description: 'document_remove_site_success_response', )]
     #[IdParameter(type: ElementTypes::TYPE_DOCUMENT)]
     #[DefaultResponses([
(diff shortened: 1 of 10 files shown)


src/Document/Controller/Site/DeleteController.php
```diff
@@ -0,0 +1,73 @@
...
+        private readonly SiteServiceInterface $siteService,
+        SerializerInterface $serializer,
+    ) {
+        parent::__construct($serializer);
+    }
+
+    /**
+     * @throws ForbiddenException|UserNotFoundException|NotFoundException
+     */
+    #[Route(self::ROUTE, name: 'pimcore_studio_api_documents_site_remove', methods: ['DELETE'])]
```
martineiber:
```suggestion
    #[Route(self::ROUTE, name: 'pimcore_studio_api_documents_site_delete', methods: ['DELETE'])]
```
a bit petty.. i know. :D
But we mix `remove` and `delete` here a few times. Should pick one?

martineiber (changes requested):

src/Document/Controller/Site/DeleteController.php
lukmzig: Should be adapted now :)

martineiber (approved):

PR #1095, opened 2025-05-22
[Documents][Editor] List controllers and teplates

### LINKED ISSUE #1093: [Documents][Editor] List controllers and teplates
Add endpoints to retrieve:

- list of available controllers (classic UI `/misc/get-available-controller-references`)
- list of available templates  (classic UI `/misc/get-available-templates`)

lukmzig:
Resolves #1093

### DIFF
--- a/src/Document/Controller/PageSnippet/ListControllersController.php
+++ b/src/Document/Controller/PageSnippet/ListControllersController.php
@@ -1,0 +1,72 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * This source file is available under the terms of the
+ * Pimcore Open Core License (POCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (https://www.pimcore.com)
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Document\Controller\PageSnippet;
+
+use OpenApi\Attributes\Get;
+use Pimcore\Bundle\StudioBackendBundle\Controller\AbstractApiController;
+use Pimcore\Bundle\StudioBackendBundle\Document\Schema\PageSnippet\Controller;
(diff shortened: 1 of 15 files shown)


src/Document/Controller/PageSnippet/ListControllersController.php
```diff
@@ -0,0 +1,72 @@
...
+use Pimcore\Bundle\StudioBackendBundle\OpenApi\Config\Tags;
+use Pimcore\Bundle\StudioBackendBundle\Util\Constant\HttpResponseCodes;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\Routing\Attribute\Route;
+use Symfony\Component\Serializer\SerializerInterface;
+
+/**
+ * @internal
+ */
+final class ListControllersController extends AbstractApiController
```
martineiber:
```suggestion
final class GetAvailableControllersController extends AbstractApiController
```

src/Document/Controller/PageSnippet/ListTemplatesController.php
```diff
@@ -0,0 +1,71 @@
...
+use Pimcore\Bundle\StudioBackendBundle\OpenApi\Config\Tags;
+use Pimcore\Bundle\StudioBackendBundle\Util\Constant\HttpResponseCodes;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\Routing\Attribute\Route;
+use Symfony\Component\Serializer\SerializerInterface;
+
+/**
+ * @internal
+ */
+final class ListTemplatesController extends AbstractApiController
```
martineiber:
```suggestion
final class GetAvailableTemplatesController extends AbstractApiController
```

martineiber (changes requested):
Just an Idea. So it would be easier to find the right controller when just looking at the endpoint. 

I general there is `list` and `get` mixed up some times?

lukmzig:
> Just an Idea. So it would be easier to find the right controller when just looking at the endpoint.
> 
> I general there is `list` and `get` mixed up some times?

Well we use `get` mostly for single results, this is a list. But I agree this is somewhat a special case and `get` would fit more here

martineiber (approved):

PR #1098, opened 2025-05-22
[Data objects] localized fields are empty

lukmzig:
Pat of https://github.com/pimcore/studio-ui-bundle/issues/1525

## Additional info
When user has no restricted languages in view permissions, consider it as all of them are allowed (same as in the Classic UI)

### DIFF
--- a/src/User/Schema/User.php
+++ b/src/User/Schema/User.php
@@ -50,7 +50,7 @@ final class User implements AdditionalAttributesInterface
         #[Property(description: 'If a User is active', type: 'boolean', example: true)]
         private readonly bool $active,
         #[Property(description: 'If User is admin', type: 'boolean', example: false)]
-        private bool $admin,
+        private readonly bool $admin,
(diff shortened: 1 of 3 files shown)


src/User/Schema/User.php
```diff
@@ -50,7 +50,7 @@ public function __construct(
         #[Property(description: 'If a User is active', type: 'boolean', example: true)]
         private readonly bool $active,
         #[Property(description: 'If User is admin', type: 'boolean', example: false)]
-        private bool $admin,
+        private readonly bool $admin,
```
martineiber, line 53: Let's make the hole class Readonly?

martineiber (changes requested):

src/User/Schema/User.php
martineiber: Sorry my mistake. We can not do that b/c of AdditionalAttributesTrait

martineiber (approved):

PR #1109, opened 2025-05-26
[Translations] Only show public translation when not logged in.

### LINKED ISSUE #1107: Wrong return when requesting translations
I discovered that you get all translations if you are not logged in and send an empty keys array. 
This shouldn`t be the case ...

As discussed, we should return all public translations of you trigger the endpoint (without a session).

martineiber:
Resolves #1107

### DIFF
--- a/src/Security/Service/SecurityService.php
+++ b/src/Security/Service/SecurityService.php
@@ -47,6 +47,16 @@ final readonly class SecurityService implements SecurityServiceInterface
         return $pimcoreUser;
     }
 
+    public function isLoggedIn(): bool
+    {
+        try {
+            $this->getCurrentUser();
+            return true;
+        } catch (UserNotFoundException) {
+            return false;
+        }
+    }
+
     /**
      * @throws ForbiddenException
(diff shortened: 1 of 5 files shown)


lukmzig (changes requested):
https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Security/Voter/PublicAuthorizationVoter.php#L58

I think we should replace this also with the new public method `isLoggedIn` from `SecurityServiceInterface` to avoid duplications

martineiber:
> https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Security/Voter/PublicAuthorizationVoter.php#L58
> 
> I think we should replace this also with the new public method `isLoggedIn` from `SecurityServiceInterface` to avoid duplications

Good Point! I have changed this here [fd056f7](https://github.com/pimcore/studio-backend-bundle/pull/1109/commits/fd056f74ac2096af90036f207825fed021614df5)

lukmzig (approved):

PR #1131, opened 2025-06-10
[Filter] Enable Date Filter for DataObjects and use ISO 8601

### LINKED ISSUE #1123: Ignore time value when filtering with "on" date filter
See related task: https://github.com/pimcore/studio-ui-bundle/issues/1514 and comments.

We've decided to remove the time options from the date picker. Could you please align the backend accordingly? When I filter by a specific date (without time), I should receive all data matching that date. Currently, that doesn't work.

**This is relevant for assets and data objects.**

Image
Image

martineiber:
Resolves #1123

## Additional info
Date Filter now requires ISO 8601 Time. This includes the Timezone. 

Needs https://github.com/pimcore/generic-data-index-bundle/pull/334


lukmzig (approved):
LGTM

PR #1153, opened 2025-06-17
[Task] Advanced Columns improvement

martineiber:
Part of https://github.com/pimcore/studio-ui-bundle/issues/299
## Additional info

- Support all data types
- Concatenation is part of the Transformer.


src/Grid/Column/Transformer/Combine.php
```diff
@@ -0,0 +1,68 @@
...
+use function is_string;
+
+/**
+ * @internal
+ */
+final class Combine implements TransformerInterface
+{
+    public function transform(array $value, array $config): array
+    {
+        if (!isset($config['clue'])) {
```
lukmzig: I am not completely sure, but should this not be "glue"?

martineiber: Good Point Thx :)

martineiber: fixed it here [1c3d8ab](https://github.com/pimcore/studio-backend-bundle/pull/1153/commits/1c3d8abe5b64e017735b87c922354901d6c9db55)

lukmzig (approved):
LGTM

PR #1168, opened 2025-06-26
[Application Logger]: list log entries

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/1644

## Additional info
Add endpoint and filters to list log entries

### DIFF
--- a/src/Bundle/ApplicationLogger/Controller/CollectionController.php
+++ b/src/Bundle/ApplicationLogger/Controller/CollectionController.php
@@ -17,6 +17,7 @@ use OpenApi\Attributes\Post;
 use Pimcore\Bundle\StudioBackendBundle\Bundle\ApplicationLogger\Schema\LogEntry;
 use Pimcore\Bundle\StudioBackendBundle\Bundle\ApplicationLogger\Service\LogServiceInterface;
 use Pimcore\Bundle\StudioBackendBundle\Controller\AbstractApiController;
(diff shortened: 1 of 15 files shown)


src/Bundle/ApplicationLogger/Controller/CollectionController.php
```diff
@@ -0,0 +1,97 @@
...
+            . ']',
+        sortFilterExample: '{"key":"id", "direction":"DESC"}'
+    )]
+    #[SuccessResponse(
+        description: 'bundle_application_logger_get_collection_success_response',
+        content: new CollectionJson(new GenericCollection(LogEntry::class))
+    )]
+    #[DefaultResponses([
+        HttpResponseCodes::UNAUTHORIZED,
+        HttpResponseCodes::NOT_FOUND,
```
martineiber: I think this one can be removed?

martineiber (changes requested):

src/Bundle/ApplicationLogger/Controller/CollectionController.php
lukmzig: Thanks :) updated the responses here: https://github.com/pimcore/studio-backend-bundle/pull/1168/commits/10bae618fb80c166367bcad632cff062728f8f3d

martineiber (approved):
LGTM

PR #1169, opened 2025-06-26
[Application Logger]: Add bundle loading

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/1644

## Additional info
Load services and open api configuration for bundle only if its registered and installed

### DIFF
--- a/src/OpenApi/Service/OpenApiService.php
+++ b/src/OpenApi/Service/OpenApiService.php
@@ -20,13 +20,13 @@ use OpenApi\Annotations\Tag;
 use OpenApi\Attributes\Schema;
 use OpenApi\Attributes\Server;
 use OpenApi\Generator;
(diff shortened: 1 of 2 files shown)


src/OpenApi/Service/OpenApiService.php
```diff
@@ -143,4 +151,56 @@ private function checkValidOpenApiScanPaths(): void
...
+            }
+        }
+
+        $config->components->schemas = array_values($config->components->schemas);
+
+        $config->tags = array_filter(
+            $config->tags,
+            static fn (Tag $tag) => $tag->name !== $tagName
+        );
+        $config->tags = array_values($config->tags);
```
martineiber: Can we move this into the catch block, and remove the return in line 181? I I think then this is a bit better to read?

martineiber (changes requested):

martineiber (approved):
LGTM

PR #1176, opened 2025-06-30
[Redirects] Add endpoint to list and filter redirects

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/1645

## Additional info
- add listing endpoint
- add filtering and search

### DIFF
--- a/src/Bundle/Seo/Event/PreResponse/RedirectListEvent.php
+++ b/src/Bundle/Seo/Event/PreResponse/RedirectListEvent.php
@@ -1,0 +1,36 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * This source file is available under the terms of the
+ * Pimcore Open Core License (POCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (https://www.pimcore.com)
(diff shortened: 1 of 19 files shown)


src/Bundle/Seo/Event/PreResponse/RedirectListEvent.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Bundle\Seo\Event\PreResponse;
+
+use Pimcore\Bundle\StudioBackendBundle\Bundle\Seo\Schema\Redirect;
+use Pimcore\Bundle\StudioBackendBundle\Event\AbstractPreResponseEvent;
+
+final class RedirectListEvent extends AbstractPreResponseEvent
+{
+    public const string EVENT_NAME = 'pre_response.bundle_seo.redirect.list';
```
martineiber, line 21: Can you add this event to the docs?

martineiber (changes requested):

src/Bundle/Seo/Event/PreResponse/RedirectListEvent.php
lukmzig: good point, thanks :)

martineiber (approved):

PR #1177, opened 2025-07-01
Update ContextPermissionService.php

Corepex:
Hi Team, 

according to: 
https://github.com/pimcore/pimcore/blob/f2ec89bb94fdf6fd2bca66472a7b179b69164269/bundles/ApplicationLoggerBundle/public/js/startup.js#L26

the permission should be named `applicationlog`.  (not camelcase).

## Additional info

This pull request includes a minor adjustment to the `ContextPermissionService` class. The change corrects the casing of the `applicationLog` key in the `$extraPermissions` array to ensure consistency with other keys.


martineiber (approved):
LGTM

PR #1182, opened 2025-07-02
[Website Settings]: Update list data schema

### LINKED ISSUE #1180: [Website Settings] - please return element fullPath and Id to avoid unecessary further fetching in frontend
https://studionightly.pimcore.app/pimcore-studio/api/docs#/Website%20Settings/website_settings_get_collection

As part of the website settings, the backend is currently returning the fullPath of an element inside "data"

As far as I am informed there is a convention to always send the fullpath and element Id together.
This makes it easier to display the element in the FE (therefore we dont have to make a second api call to get the ID)
 
in the link below you can see how we re also receiving both on for example NotificationsDetails
 
https://studionightly.pimcore.app/pimcore-studio/api/docs#/Notifications/notification_get_by_id

In the FE when we try to update the data field we change the "fullPath" "id" etc we get returned from the component to be a string inside "data" for the call to work.
I wonder if we should update this too. 

below the shape of what our element field returns atm. 

export interface ElementInfo {
  elementType?: ElementType
  id?: number
  fullPath: string | false
  published?: boolean
  disabled?: boolean
}

Get in touch if you would like to discuss this further. Thanks

lukmzig:
Resolves #1180 

## Additional info
- FE requires ID when element is passed as data
- update data schema to reflect all possible return types

### DIFF
--- a/src/WebsiteSetting/Hydrator/WebsiteSettingsHydrator.php
+++ b/src/WebsiteSetting/Hydrator/WebsiteSettingsHydrator.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Pimcore\Bundle\StudioBackendBundle\WebsiteSetting\Hydrator;
 
 use Pimcore\Bundle\StudioBackendBundle\Util\Constant\ElementTypes;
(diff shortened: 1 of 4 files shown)


martineiber (approved):
LGTM

PR #1192, opened 2025-07-07
[Reports] Add drill down functionality

### LINKED ISSUE #1185: [Reports] Add Drilldown filter functionality
In the old admin in Report, we have a drilldown filter with the following options:
Image

We get this data from the following endpoint:
Image

Need to add this functionality to the API. Also, please add information in the comment about the approach to retrieving the drilldown label.
Image

lukmzig:
Resolves #1185

## Additional info
- add drill down filter info to configuration endpoint
- add drill down filter options endpoint


martineiber (approved):
LGTM

PR #1216, opened 2025-07-14
[Translation] Add Domain to Create/Delete/Update

### LINKED ISSUE #1191: [Translations - Edit Mode] - new API call
Image

As discussed with Martin. 
Please add a classic get collection API call that allows us to get a translation key with its translations for the locales, type, creationDate, modificationDate.

Please use the shape: 
{
totalItems: x,
items: x
}

above foto of admin api call response for the same use case

martineiber:
Resolves #1191


lukmzig (approved):
LGTM

PR #1220, opened 2025-07-15
[Translation] Add search filter

### LINKED ISSUE #1219: [Translation] Add overall search Filter
Search Filter is missing. 

We need a search filter for key and all translations. 

Image

martineiber:
Resolves #1219

## Additional info
Search filter can be applied like that:
```json
{
  "type": "search",
  "filterValue": "Accessory"
}
```

### DIFF
--- a/src/Translation/Service/TranslatorService.php
+++ b/src/Translation/Service/TranslatorService.php
@@ -34,6 +34,7 @@ use Pimcore\Translation\Translator;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Translation\TranslatorBagInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
+use function array_key_exists;
 use function in_array;
 
 /**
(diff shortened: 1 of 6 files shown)


src/Translation/Service/TranslatorService.php
```diff
@@ -151,13 +152,23 @@ public function listTranslations(string $domain, CollectionFilterParameter $para
...
+                !in_array($columnFilter['key'], $validLanguages, true)
+            ) {
                 continue;
             }
             $joins[] = $columnFilter['key'];
         }
 
         $this->translationRepository->joinLanguageColumns($list, $joins, $domain);
+
+        $searchFilter = $parameter->getFilters()->getSimpleColumnFilterByType('search');
```
lukmzig: What do you think if we would introduce a new `search` type here: https://github.com/pimcore/studio-backend-bundle/blob/1.x/src/Filter/FilterType.php

so we could re-use it then in other listings?

lukmzig (approved):

PR #1232, opened 2025-07-22
[RecycleBin] Rename perspective permission name

Corepex:
## Additional info

According to the admin-ui the perspective permission should be named `recyclebin` instead of `recycle_bin`

https://github.com/pimcore/admin-ui-classic-bundle/blob/7bf40f83bc1e93bd58decc2db98a85015a5d8461/public/js/pimcore/layout/toolbar.js#L316C83-L316C93

Dont know if that's all it takes to rename the permission. Please double check ....


lukmzig:
@Corepex perspective permisisons are introduced as new for studio, they dont have to be same as in classic ui ...

perspectives and widgets are introduced completely from scratch for studio

Corepex:
not needed as it seems

PR #1262, opened 2025-08-06
[Language] Add allowed Languages For View/Edit to current user information

### LINKED ISSUE #1207: [Languages] - Endpoint to relate availableLanguages i.e. "en" to a language ie. "English"
This could just give me all the available languages and their "display name" / or it could give a list of keys i.e. "en" with display values "English". 

Lets discuss / see what its like in the classic ui

martineiber:
Resolves #1207


lukmzig (changes requested):
I think in schema we need to update required fields 😬

martineiber:
> I think in schema we need to update required fields 😬

Thx.. 
Fixed it

Issue #1245, opened 2025-07-29
[Assets] Add image thumbnail option to custom image stream endpoint


markus-moser:
Add an option to define a thumbnail definition to the `/pimcore-studio/api/assets/{id}/image/stream/custom` endpoint.

We need this for the document editables to create a pendant to this request: `/admin/asset/get-image-thumbnail?id=337&thumbnail=portalCarousel`

lukmzig:
@markus-moser I added a new endpoint `/pimcore-studio/api/assets/{id}/video/stream/{thumbnailName}` so we do not mix custom thumbnail with pre-defined ones

PR #1284, opened 2025-08-14
[Assets] Add document custom thumbnail download and stream endpoints

### LINKED ISSUE #1257: [Assets] Add document image thumbnail endpoint
We need a pendant for this endpoint `/admin/asset/get-document-thumbnail?id=2004&pageCount=1&width=640&aspectratio=true``

Can we do it in a way so that it supports the same options like our `/pimcore-studio/api/assets/{id}/image/download/custom` endpoint but just for document assets?

lukmzig:
Resolves #1257

## Additional info
- add endpoint to download custom thumbnail for document assets
- add endpoint to stream custom thumbnail for document assets
- add endpoint to stream thumbnail by name for document assets
- add endpoint to download thumbnail by name for document assets


lukmzig:
@markus-moser could you have a quick look if this is sufficient for the requirements? I added 2 endpoints, so we are able to stream and also download custom thumbnail image for documents

markus-moser:
@lukmzig Thanks, this is exactly what we need! The only thing whats missing is that we also need the thumbnail name option like here: https://github.com/pimcore/studio-backend-bundle/issues/1245 Could you please add this endpoint too?

martineiber (approved):
LGTM

PR #1299, opened 2025-08-21
[Login]: Forgot Password

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/256

## Additional info
- addapt endpoint for reset password mail generation
- add new endpoint to log in user by token

### DIFF
--- a/src/Security/Authenticator/AdminTokenAuthenticator.php
+++ b/src/Security/Authenticator/AdminTokenAuthenticator.php
@@ -54,7 +54,7 @@ class AdminTokenAuthenticator extends AbstractAuthenticator
         );
 
         if ($pimcoreUser === null) {
-            throw new AccessDeniedException('Failed to authenticate with username and token');
+            throw new AccessDeniedException('Failed to authenticate with the given token');
         }
 
         $pimcoreUser->setTwoFactorAuthentication('required', false);
--- a/src/Security/Authenticator/AdminTokenAuthenticator.php
+++ b/src/Security/Authenticator/AdminTokenAuthenticator.php
@@ -1,0 +1,111 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * This source file is available under the terms of the
+ * Pimcore Open Core License (POCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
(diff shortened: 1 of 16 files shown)


src/Security/Authenticator/AdminTokenAuthenticator.php
```diff
@@ -0,0 +1,113 @@
...
+        return ($request->attributes->get('_route') === TokenLoginController::ROUTE_NAME) &&
+            $this->getTokenFromRequest($request) !== null;
+    }
+
+    /**
+     *  @throws AccessDeniedException|InvalidArgumentException
+     */
+    public function authenticate(Request $request): Passport
+    {
+        $pimcoreUser = Authentication::authenticateToken(
```
martineiber, line 52: Just to be sure. This also invalidates the token? The token should only be valid once?

src/Security/Authenticator/AdminTokenAuthenticator.php
```diff
@@ -0,0 +1,113 @@
...
+            $this->getTokenFromRequest($request),
+        );
+
+        if ($pimcoreUser === null) {
+            throw new AccessDeniedException('Failed to authenticate with username and token');
```
martineiber:
```suggestion
            throw new AccessDeniedException('Failed to authenticate with the given token');
```

We only provide the token and not a username?

src/Security/Authenticator/AdminTokenAuthenticator.php
```diff
@@ -0,0 +1,113 @@
...
+        return new SelfValidatingPassport($userBadge);
+
+    }
+
+    /**
+     * @throws AccessDeniedException
+     */
+    public function onAuthenticationFailure(Request $request, AuthenticationException $exception): ?Response
+    {
+        throw new AccessDeniedException('Failed to authenticate with username and token');
```
martineiber:
```suggestion
        throw new AccessDeniedException('Failed to authenticate with the given token');
```

martineiber (changes requested):
Just a small suggestion.

src/Security/Authenticator/AdminTokenAuthenticator.php
lukmzig: Yep the token is processed by core

martineiber (approved):

Issue #1233, opened 2025-07-22
[Grid Columns] Deliver all available Object Bricks


martineiber:
Make sure this also work if one class has the same Object Brick more than one time. And also adopt available-columns endpoint. 

Object Bricks should be visible in the Simple Field as Source Field form the Advanced Columns. 

The group attribute will be an array of string instate of just a string. 

Example:

Attributes > Attributes > Bodywork > Number of Doors
Attributes > Attributes > Bodywork > Number of Seats

PR #1307, opened 2025-08-26
[Grid] List all Object Brick for Advanced Columns.

martineiber:
Part of #1233


lukmzig (changes requested):
Hmm can we move this `Pimcore\Bundle\StudioBackendBundle\ObjectBrick\Service\ObjectBrickService` rather in the DataObject namespace? Maybe `DataObejct\Service\ObjectBrick`?

martineiber:
> Hmm can we move this `Pimcore\Bundle\StudioBackendBundle\ObjectBrick\Service\ObjectBrickService` rather in the DataObject namespace? Maybe `DataObejct\Service\ObjectBrick`?

The idea was that, if we add more functionality for ObjectBricks in the future—such as list, updates and similar features—it will all stay within the ObjectBrick namespace, just like we already do for the ClassificationStore.

WDYT?

lukmzig:
> > Hmm can we move this `Pimcore\Bundle\StudioBackendBundle\ObjectBrick\Service\ObjectBrickService` rather in the DataObject namespace? Maybe `DataObejct\Service\ObjectBrick`?
> 
> The idea was that, if we add more functionality for ObjectBricks in the future—such as list, updates and similar features—it will all stay within the ObjectBrick namespace, just like we already do for the ClassificationStore.
> 
> WDYT?

Ah I see ..in that case you are right, the separate namespace is better

lukmzig (approved):

PR #1312, opened 2025-08-27
[Translations] - API call for translation import

### LINKED ISSUE #1205: [Translations] - API calls for CSV download, import and merge for Translations as in admin UI
as in admin ui

### LINKED ISSUE #1285: [Translations] allow updating type in relation to key not locale translation
You can see below that all translation updates are made focussing around a specific locale i.e. 'en'. The type update is in reality across all locales. 
 
With the current API set up, this means that when updating the type I have to use an arbitrary locale (I am using the first in line) and send the updated type nested inside the locale. 

It would make sense to allow updating the type based on the key without having to add any translations in the update.  

Image

lukmzig:
Resolves #1205 
Resolves https://github.com/pimcore/studio-backend-bundle/issues/1285

## Additional info
- add endpoint to get csv settings based on file content
- add endpoint to import translations and calculate delta
- adjust update endpoint to allow batch update and update by key rather than locale


martineiber (approved):
LGTM

PR #1341, opened 2025-09-09
[Assets] Add crop options to all image thumbnail stream endpoints

### LINKED ISSUE #1292: [Assets] Custom thumbnail option in stream endpoints
I found another thumbnail option which I have forgotten to mention:
https://docs.pimcore.com/platform/Pimcore/Documents/Editables/Image#an-example-with-a-direct-thumbnail-configuration

There it's possible to define a custom thumbnail inline. In the classic bundle this results in a URL like this:
`/admin/asset/get-image-thumbnail?id=1771&alt=&cropPercent=false&cropWidth=0&cropHeight=0&cropTop=0&cropLeft=0&thumbnail%5Bwidth%5D=200&thumbnail%5Bheight%5D=200&thumbnail%5Binterlace%5D=true&thumbnail%5Bquality%5D=90&height=undefined&pimcore_editmode=1`

 Can we somehow add this for images and documents? Maybe with a JSON parameter in our custom stream endpoints?

### LINKED ISSUE #1329: [Assets] Add crop options to all image thumbnail stream endpoints
We also need the crop parameters also for our thumbnail based endpoints:

- /pimcore-studio/api/assets/{id}/document/stream/thumbnail/{thumbnailName}
- /pimcore-studio/api/assets/{id}/image/stream/thumbnail/{thumbnailName}

And also we need to respect it there:
https://github.com/pimcore/studio-backend-bundle/issues/1292

lukmzig:
Resolves #1329
Resolves #1292

## Additional info
- add endpoints for image and document image streams with dynamic configuration (config creation is done by core)
- add possibility to crop named thumbnails


src/Asset/Attribute/Property/MimeType.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+final class MimeType extends Property
+{
+    public function __construct()
+    {
+        parent::__construct(
+            'mimeType',
+            description: 'Define format of the asset thumbnail',
+            type: 'string',
+            nullable: true
```
martineiber: Can we add an example?

martineiber (changes requested):

src/Asset/Attribute/Property/MimeType.php
lukmzig: I removed it because its actually not used anymore

martineiber (approved):

PR #1343, opened 2025-09-09
[User Management] Two factor option

### LINKED ISSUE #1234: [User Management] Two factor option
Currently the twoFactorAuthenticationEnabled option is always set to false, even when it is activated in the classic bundle.

Image

Image

lukmzig:
Resolves #1234

## Additional info
Add missing two factor authentication data


martineiber (approved):
LGTM

PR #1344, opened 2025-09-10
[User management] delete avatar endpoint missing

### LINKED ISSUE #1235: [User management] delete avatar endpoint missing
We need an endpoint to delete the user's avatar image.

example from classic-ui: 

Image

### LINKED ISSUE #1237: [Profile] Avatar image profile permissions
The currently logged-in user doesn't have permission to change their own avatar image using the upload-image endpoint. Maybe we need an additional endpoint, such as update-profile, or should it be possible to use the upload-image endpoint?

lukmzig:
Resolves #1235 
Resolves #1237

## Additional info
- adapter user permission checks for avatar endpoints since they are shared with profile page
- add endpoint to remove user avatar


vin0401:
@lukmzig - I came across some issue with the avatar as well.
The user needs permission to request his own avatar kind of. :D 

image

Would that be also handled with this PR?

lukmzig:
> @lukmzig - I came across some issue with the avatar as well. The user needs permission to request his own avatar kind of. :D
> 
> image
> Would that be also handled with this PR?

Thanks for the info. I adjusted only update and delete so far but I will adapt the `GET` request as well.

src/User/Service/ImageService.php
```diff
@@ -34,17 +35,86 @@ public function __construct(
...
+     * @throws ForbiddenException
+     */
+    private function validateCurrentUser(UserInterface $user, UserInterface $currentUser): void
+    {
+        if (!$currentUser->isAllowed(UserPermissions::USER_MANAGEMENT->value) &&
+            ($user->getId() !== $currentUser->getId())
+        ) {
+            throw new ForbiddenException('You do not have required permissions for this action');
+        }
+    }
```
martineiber, line 114: Would it make sense to move this to a Voter? 

WDYT?

martineiber (changes requested):

src/User/Service/ImageService.php
lukmzig: I am not really sure why? We compare queried user with current user and also we have different case for getting an image and other operations, so we would need different voters for different actions.

martineiber (approved):

Issue #1155, opened 2025-06-17
[Grid] [Advanced Columns] Add Transformers


martineiber:
- [x] twig operator
    - Should take the normalized value
    - Input will be the twig template
    - Frontend will have a Code Editor. 
- [x] anonymizer
- [x] element counter
- [x] concatenator
- [x] string replace
- [x] substring
- [x] translate
   - See https://github.com/pimcore/admin-ui-classic-bundle/blob/10ea29130a41a9fefbcfc57b082f0ec5e3637a9b/src/DataObject/GridColumnConfig/Operator/TranslateValue.php
- [x] trim
- [x] php code
    - Tagged Service. This one can be selected executed in the Transformer.  
    - Interface
    - Endpoint to list all tagged Services. 
- [x] boolean formatter
- [x] dateformatter
- [x] explode

PR #1382, opened 2025-09-18
[Transformer] Add basic Transformer

stunnerparas:
Part of #1155


src/Grid/Column/Transformer/BooleanFormatter.php
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        $trueLabel = $config['trueLabel'];
+        $falseLabel = $config['falseLabel'];
+        $results = [];
+
+        foreach ($value as $val) {
+            $data = $val->getValue();
+            if (!is_bool($data)) {
+                $results[] = new AdvancedValue('string', '');
```
martineiber: I would say we return the original Value!

src/Grid/Column/Transformer/Concatenator.php
```diff
@@ -0,0 +1,77 @@
...
+use Pimcore\Bundle\StudioBackendBundle\Grid\Column\TransformerInterface;
+use Pimcore\Bundle\StudioBackendBundle\Grid\Util\AdvancedValue;
+use function array_map;
+use function implode;
+use function is_string;
+
+/**
+ * @internal
+ */
+final class Concatenator implements TransformerInterface
```
martineiber: This one it not needed it is the same as the Combine Transform which already exists.

src/Grid/Column/Transformer/StringReplace.php
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        $find = $config['find'];
+        $replace = $config['replace'];
+        $results = [];
+
+        foreach ($value as $val) {
+            $data = $val->getValue();
+            if (!is_string($data)) {
+                $results[] = new AdvancedValue('string', $data);
```
martineiber: The type should not be string it should be the original type when we return the original value.

src/Grid/Column/Transformer/SubString.php
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        $start = $config['start'];
+        $length = $config['length'];
+        $results = [];
+
+        foreach ($value as $val) {
+            $data = $val->getValue();
+            if (!is_string($data)) {
+                $results[] = new AdvancedValue('string', $data);
```
martineiber: The type should not be string it should be the original type when we return the original value.

src/Grid/Column/Transformer/Trim.php
```diff
@@ -0,0 +1,63 @@
...
+final class Trim implements TransformerInterface
+{
+    public function transform(array $value, array $config): array
+    {
+        $results = [];
+
+        foreach ($value as $val) {
+            $data = $val->getValue();
+            if (!is_string($data)) {
+                $results[] = new AdvancedValue('string', $data);
```
martineiber: The type should not be string it should be the original type when we return the original value.

src/Grid/Column/Transformer/BooleanFormatter.php
```diff
@@ -0,0 +1,84 @@
...
+ * @internal
+ */
+final class BooleanFormatter implements TransformerInterface
+{
+    public function transform(array $value, array $config): array
+    {
+        if (!isset($config['trueLabel']) || !is_string($config['trueLabel'])) {
+            throw new TransformerException(
+                $this->getName(),
+                sprintf('Missing or invalid "trueLabel" configuration for %s transformer.', $this->getKey())
```
martineiber: Can we add to all checks what type the configuration should be to the Exception Message. In that case, that it must be a string.

martineiber (changes requested):
Some small changes.

src/Grid/Column/Transformer/Trim.php
```diff
@@ -0,0 +1,63 @@
...
+
+use Pimcore\Bundle\StudioBackendBundle\Grid\Column\TransformerInterface;
+use Pimcore\Bundle\StudioBackendBundle\Grid\Util\AdvancedValue;
+use function is_string;
+use function trim;
+
+/**
+ * @internal
+ */
+final class Trim implements TransformerInterface
```
martineiber, line 26: Trim should have this options. 
image

src/Grid/Column/Transformer/BooleanFormatter.php
```diff
@@ -0,0 +1,91 @@
...
+
+        $trueLabel = $config['trueLabel'];
+        $falseLabel = $config['falseLabel'];
+        $results = [];
+
+        foreach ($value as $val) {
+            $data = $val->getValue();
+            if (!is_bool($data)) {
+                $originalType = $val->getType();
+                $results[] = new AdvancedValue($originalType, $data);
```
martineiber: $originalValue is useless here. 

```suggestion
                $results[] = new AdvancedValue($$val->getType(), $data);
```

martineiber (changes requested):
One more small change.

martineiber (approved):
LGTM

Issue #1331, opened 2025-09-02
[Data Object][Grid] Add endpoints to integrate classification store columns and filters


markus-moser:
- [x] Get Data in Grid
- [x] Endpoint to get Group ID and Key ID for Data Endpoint (Modal)
- [x] Filters

martineiber:
See https://github.com/pimcore/studio-ui-bundle/issues/905

PR #1401, opened 2025-09-24
[Grid] Add Multi Select and Multi Boolean Select

### LINKED ISSUE #1365: [Grid][Filter][Multiselect] - filtering for an array of strings does not seem to exist
Image

same applies for boolean-select i.e. an array with possible values: false, null, true
https://github.com/pimcore/studio-backend-bundle/blob/1.x/doc/03_Grid.md#filter

I have used metadata.select which does not give me an error but also does not return a filtered result

martineiber:
Resolves #1365
Part of #1331

## Additional info
Also for Classification Store

### DIFF
--- a/doc/03_Grid.md
+++ b/doc/03_Grid.md
@@ -49,7 +49,7 @@ Available filters are:
 | system.fulltext                          | string             |                                    | false          |
 | system.boolean                           | array              | `true`, `false` or `null`          | true           |
 | system.number                            | object             | `from`, `to`, `is`, `setting`      | true           |
-| system.select                            | srray              |                                    | true           |
+| system.select                            | array              |                                    | true           |
 | classificationstore.string               | string             |                                    | true           |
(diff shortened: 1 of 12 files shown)


doc/03_Grid.md
```diff
@@ -47,14 +47,16 @@ Available filters are:
 | system.ids                               | array of integer   |                                    | false          |
 | system.integer                           | integer            |                                    | true           |
 | system.fulltext                          | string             |                                    | false          |
-| system.boolean                           | boolean or null    |                                    | true           |
+| system.boolean                           | array              | `true`, `false` or `null`          | true           |
 | system.number                            | object             | `from`, `to`, `is`, `setting`      | true           |
+| system.select                            | srray              |                                    | true           |
```
lukmzig:
```suggestion
| system.select                            | array              |                                    | true           |
```

martineiber: thx :)

lukmzig (approved):

PR #1414, opened 2025-09-26
1155

stunnerparas:
## Twig Operator as a Transformer


martineiber:
See https://github.com/pimcore/studio-backend-bundle/pull/1417

PR #1417, opened 2025-09-26
[Grid] [Transformer] Twig Transformer

stunnerparas:
Part of #1155

### DIFF
--- a/src/Grid/Column/Transformer/TwigOperator.php
+++ b/src/Grid/Column/Transformer/TwigOperator.php
@@ -29,26 +29,39 @@ final class TwigOperator implements TransformerInterface
 
     public function transform(array $value, array $config): array
     {
+        if (isset($config['template']) && !is_string($config['template'])) {
+            throw new TransformerException(
+                $this->getName(),
+                sprintf(
+                    'Invalid "template" configuration (must be a string) for %s transformer. ' .
+                    'Example: "template": "{{ value|date(\'d.m.Y H:i\') }}"',
+                    $this->getKey()
+                )
+            );
+        }
+
         $template = $config['template'] ?? '{{ value }}';
 
+        $results = [];
+
         foreach ($value as $val) {
-            try {
-                $rendered = $this->templateGenerator->generate($template, ['value' => $val->getValue()]);
-                $val->setValue($rendered);
-            } catch (Throwable $e) {
+            $rendered = $this->templateGenerator->generate($template, ['value' => $val->getValue()]);
+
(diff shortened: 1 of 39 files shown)


src/Grid/Column/Transformer/TwigOperator.php
```diff
@@ -0,0 +1,81 @@
...
+
+    public function transform(array $value, array $config): array
+    {
+        $template = $config['template'] ?? '{{ value }}';
+
+        foreach ($value as $val) {
+            try {
+                $rendered = $this->templateGenerator->generate($template, ['value' => $val->getValue()]);
+                $val->setValue($rendered);
+            } catch (Throwable $e) {
```
martineiber: Only catch `Throwable` if it is necessary for the logic. This hiddes a lot of errors. Better use `Exception` or more specific.

src/Twig/TemplateGenerator.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public function generate(string $twigTemplate, array $arguments): string
+    {
+        try {
+            $this->sandboxExtension->enableSandbox();
+
+            return $this->twig->createTemplate($twigTemplate)->render($arguments);
+        } catch (Throwable $e) {
+            throw new TransformerException(
```
martineiber: `TransformerException` isn’t related to the Twig Sandbox, so it doesn’t seem like the right exception to throw here. Also, keep in mind that this Twig generator may be used in other parts of the code.


Also, avoid caching Throwable.
Twig’s base error class already extends Exception:
https://github.com/twigphp/Twig/blob/3.x/src/Error/Error.php#L36

src/Grid/Column/Transformer/TwigOperator.php
```diff
@@ -0,0 +1,81 @@
...
+final class TwigOperator implements TransformerInterface
+{
+    public function __construct(
+        private readonly TemplateGeneratorInterface $templateGenerator
+    ) {
+    }
+
+    public function transform(array $value, array $config): array
+    {
+        $template = $config['template'] ?? '{{ value }}';
```
martineiber, line 43: Please also update the documentation with a small example, so users can see how to access the data variable in Twig.

https://github.com/pimcore/studio-backend-bundle/blob/1.x/doc/03_Grid.md#transformers

martineiber (changes requested):
Some small changes.

martineiber:
See https://github.com/pimcore/studio-backend-bundle/pull/1429

PR #1423, opened 2025-09-29
[User management]: handle missing properties

### LINKED ISSUE #1386: [User Management] Date Time Locale is missing
dateTimeLocale value is missing in the user information (GET /pimcore-studio/api/user/{id}), even though I am sending it with the save endpoint (PUT /pimcore-studio/api/user/{id}).

Image

### LINKED ISSUE #1408: [User management]  workspaces saving
when we try to save localizedEdit, localizedView, layouts in dataObjectWorkspaces, the data is not being saved. After save the user always receives "null" as value for this options in the GET request. The same issue occurs with "save" and "unpublish"

It would also be great if we could receive the data as an array instead of a string.
e.g instead of 'de,en,fr' ['de','en','fr']

lukmzig:
Resolves https://github.com/pimcore/studio-backend-bundle/issues/1408
Resolves https://github.com/pimcore/studio-backend-bundle/issues/1386

## Additional info
- handle missing workspace data object and document properties for user update
- add missing Date time locale to response and update

### DIFF
--- a/src/Role/MappedParameter/UpdateRoleParameter.php
+++ b/src/Role/MappedParameter/UpdateRoleParameter.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace Pimcore\Bundle\StudioBackendBundle\Role\MappedParameter;
 
+use Pimcore\Bundle\StudioBackendBundle\User\Schema\UserDataObjectWorkspace;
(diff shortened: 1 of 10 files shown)


martineiber (approved):
LGTM

PR #1429, opened 2025-09-30
[Grid] [Transformer] Twig Transformer

stunnerparas:
Twig Operator Transformer with basic to general data 

Part of https://github.com/pimcore/studio-backend-bundle/issues/1155


doc/03_Grid.md
```diff
@@ -325,3 +326,56 @@ Available configurations:
...
+            { "key": "simpleField", "config": { "field": "filename" } },
+            { "key": "simpleField", "config": { "field": "classname" } },
+            { "key": "simpleField", "config": { "field": "bodyStyle" } }
+        ],
+        "transformers": [
+            {
+                "key": "twigOperator",
+                "config": {
+                    "template": "<h1>{{ value.name|trim }}</h1><p><strong>ID:</strong> {{ value.id }}</p>{% if value.color is iterable %}<p><strong>Available Colors:</strong></p><ul>{% for color in value.color %}<li>{{ color }}</li>{% endfor %}</ul>{% else %}<p><em>No colors available.</em></p>{% endif %}{% if value.bodyStyle is defined and value.bodyStyle.fullPath is defined %}<p><strong>Body Style:</strong> {{ value.bodyStyle.fullPath }}</p>{% endif %}<p><strong>Path:</strong> {{ value.fullpath }}</p><p><strong>Filename:</strong> {{ value.filename }}</p><p><strong>Class:</strong> {{ value.classname }}</p><hr><p><em>Generated summary for car object.</em></p>",
+                    "advancedColumns": [
```
martineiber: I don’t think this is necessary, as it requires passing some content twice.

I’d suggest adopt AdvancedValue here, as we already discussed, and also including the fieldname.


https://github.com/pimcore/studio-backend-bundle/blob/e9442d0d6ad592b24a4833f61a0f1aae6525989d/src/Grid/Column/Resolver/DataObject/AdvancedColumnResolver.php#L154

martineiber (changes requested):

src/Grid/Util/AdvancedValue.php
```diff
@@ -17,7 +17,8 @@ final class AdvancedValue
 {
     public function __construct(
         private string $type,
-        private mixed $value
+        private mixed $value,
+        private ?string $fieldName = null
```
martineiber: I wouldn’t make this field nullable. Each Transformer should be responsible for setting the name.
With the current approach, if another Transformer runs before the Twig Transformer, it will not work because `fieldname` is `null`, which feels unintended.

Instead, I’d suggest passing `$column->getKey()` to the Transformer through the config options. That way, each Transformer can decide how the field should be named — for example, using the column key when returning a single combined value (like `implode`), or keeping the original field key when only manipulating the existing field.

https://github.com/pimcore/studio-backend-bundle/blob/e48622df64c0d8114f4622b7ba995f0c7cb9d4e4/src/Grid/Column/Resolver/DataObject/AdvancedColumnResolver.php#L107

src/Grid/Column/Transformer/TwigOperator.php
```diff
@@ -0,0 +1,113 @@
...
+
+        if (!is_string($rendered) || trim($rendered) === '') {
+            throw new TransformerException(
+                $this->getName(),
+                sprintf(
+                    'Twig rendering did not return a valid non-empty string for %s transformer.',
+                    $this->getKey()
+                )
+            );
+        }
```
martineiber: This check is unnecessary since the return value of `generate()` is already a `string`.
I’d suggest removing it and instead handling any exceptions directly from `generate()`.

src/Twig/TemplateGeneratorInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Twig;
+
+interface TemplateGeneratorInterface
+{
+    /**
+     * @throws \Pimcore\Bundle\StudioBackendBundle\Exception\Api\TransformerException
+     */
```
martineiber, line 23: Interface does not match with the implementation.

martineiber (changes requested):

src/Grid/Util/AdvancedValue.php
```diff
@@ -17,7 +17,8 @@ final class AdvancedValue
 {
     public function __construct(
         private string $type,
-        private mixed $value
+        private mixed $value,
+        private string $fieldName
```
martineiber, line 21: Let's add a setter as well. Otherwise the property can be readonly.

src/Twig/TemplateGeneratorInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ *  @copyright  Copyright (c) Pimcore GmbH (https://www.pimcore.com)
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Twig;
+
+interface TemplateGeneratorInterface
+{
+    /**
+     * @throws \Pimcore\Bundle\StudioBackendBundle\Exception\Api\InvalidTemplateException
```
martineiber: please use `use \Pimcore\Bundle\StudioBackendBundle\Exception\Api\InvalidTemplateException`

src/Grid/Column/Transformer/TwigOperator.php
```diff
@@ -0,0 +1,105 @@
...
+
+        $template = $config['template'] ?? '{{ value }}';
+
+        $context = [
+            'value' => $this->buildAssociativeContext($value),
+        ];
+
+        $rendered = $this->templateGenerator->generate($template, $context);
```
martineiber: Catch all exception from `generate()` and  throw a TransformerException

martineiber (changes requested):
Some minor changes.

src/Grid/Column/Transformer/TwigOperator.php
```diff
@@ -0,0 +1,113 @@
...
+
+        $template = $config['template'] ?? '{{ value }}';
+
+        $context = [
+            'value' => $this->buildAssociativeContext($value),
+        ];
+
+        try {
+            $rendered = $this->templateGenerator->generate($template, $context);
+        } catch (Throwable $e) {
```
martineiber: Please catch Exception instead of Throwable. For more infos see our Coding Guidelines.

src/Grid/Util/AdvancedValue.php
martineiber: This is still missing.

martineiber (changes requested):

martineiber (approved):
LGTM

PR #1435, opened 2025-10-03
[Data Object Editor] Check size of layout JSON

### LINKED ISSUE #1036: [Data Object Editor] Check size of layout JSON
The "AllFields" class definition of this issue produces a layout endpoint file size of >2MB:
https://github.com/pimcore/studio-ui-bundle/issues/1388

`http://localhost/pimcore-studio/api/data-objects/1336/layout/0`

In the classic UI the detail endpoint (which contains laoyout + data) is only about 250KB.

`http://localhost/admin/object/get?_dc=1745335573844&id=1336&layoutId=0``

How does it happen that the new one is so much bigger?

lukmzig:
Resolves #1036 

## Additional info
Symfony Serializer was adding by default `childrenByRef` property of all children objects. In this case we dont need those and they should be skipped. This significantly reduces the response size

### DIFF
--- a/src/Controller/AbstractApiController.php
+++ b/src/Controller/AbstractApiController.php
@@ -33,9 +33,15 @@ abstract class AbstractApiController extends AbstractController
     protected function jsonResponse(
         mixed $data,
         int $status = HttpResponseCodes::SUCCESS->value,
-        array $headers = []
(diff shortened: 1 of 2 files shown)


martineiber (approved):
LGTM

PR #1439, opened 2025-10-07
[Grid] [Transformer] Anonymizer

stunnerparas:
[Grid][Transformer] Anonymizer (Mask, Initial, Partial, Hide)

Part of #1155

### DIFF
--- a/src/Grid/Column/Transformer/Anonymizer.php
+++ b/src/Grid/Column/Transformer/Anonymizer.php
@@ -51,9 +51,9 @@ final class Anonymizer implements TransformerInterface
             }
 
             $transformedValue = match ($config['rule']) {
-                'md5'    => md5($data), 
+                'md5'    => md5($data),
                 'bcrypt' => $this->bcrypt($data),
-                 default => throw new TransformerException(
+                default => throw new TransformerException(
                     $this->getName(),
                     sprintf('Invalid rule "%s" for anonymizer transformer.', $config['rule'])
                 ),
--- a/src/Grid/Column/Transformer/Anonymizer.php
+++ b/src/Grid/Column/Transformer/Anonymizer.php
@@ -18,10 +18,9 @@ use Pimcore\Bundle\StudioBackendBundle\Exception\Api\TransformerException;
 use Pimcore\Bundle\StudioBackendBundle\Grid\Column\TransformerInterface;
 use Pimcore\Bundle\StudioBackendBundle\Grid\Util\AdvancedValue;
 
-use function explode;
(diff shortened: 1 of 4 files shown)


src/Grid/Column/Transformer/Anonymizer.php
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        $results = [];
+
+        foreach ($value as $val) {
+            $data = $val->getValue();
+            $fieldName = $val->getFieldName() ?? ($config['columnKey'] ?? $this->getKey());
+
+            if (!is_string($data)) {
+                $results[] = new AdvancedValue($val->getType(), $data, $fieldName);
```
martineiber: You can just reuse the $val. Not Necessary to create it new. 

```suggestion
                $results[] = $val;
```

src/Grid/Column/Transformer/Blur.php
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        $results = [];
+
+        foreach ($value as $val) {
+            $data = $val->getValue();
+            $fieldName = $val->getFieldName() ?? ($config['columnKey'] ?? $this->getKey());
+
+            if (!is_string($data)) {
+                $results[] = new AdvancedValue($val->getType(), $data, $fieldName);
```
martineiber: You can just reuse the $val. Not Necessary to create it new. 

```suggestion
                $results[] = $val;
```

src/Grid/Column/Transformer/Anonymizer.php
```diff
@@ -0,0 +1,111 @@
...
+
+                continue;
+            }
+
+            switch ($config['rule']) {
+                case 'md5':
+                    $results[] = new AdvancedValue('string', md5($data), $fieldName);
```
martineiber: We can always here take `$val->getFieldName()`. Since we just re-map the value.

src/Grid/Column/Transformer/Blur.php
```diff
@@ -0,0 +1,187 @@
...
+                array_filter($parts)
+            )) . '.';
+        }
+
+        return implode(', ', $initials);
+    }
+
+    private function partial(string $value, array $config): string
+    {
+        $visiblePrefix = (int) ($config['visiblePrefix'] ?? 2);
```
martineiber, line 127: Please add all those available options for each mode to the Documentation.

martineiber (changes requested):

martineiber (approved):
LGTM

PR #1451, opened 2025-10-15
[Grid][Available Columns]: Add possibility to skip field types from default column processing

lukmzig:
Data object field definition fields are processed by default for all fields. We need to add possibility to skip selected fields in order to be able to add custom config and other adjustments for custom field types.


src/Grid/Column/Collector/DataObject/FieldDefinitionCollector.php
```diff
@@ -111,7 +113,9 @@ private function groupDefinitions(
             }
 
             // Use ObjectbricksCollector
```
martineiber: Let's update the comment a bit? Since it now does a bit more? 

```suggestion
            // When definition is instance of ObjectBrick there is a ObjectbricksCollector. We also skip all field types in skipFieldTypes
```

martineiber (changes requested):

martineiber (approved):

PR #1456, opened 2025-10-16
[Perspective][Widgets]: Add onlyWrapper widget property

lukmzig:
Add new property which indicates that widget is only wrapper of another configuration. 
In this case only listing and getting data via widget interface is possible, configuration data is handled via its own concrete implementation. 
This is relevant for e.g. https://github.com/pimcore/backend-power-tools-bundle/issues/489


martineiber (approved):
LGTM

PR #1520, opened 2025-11-18
[Reset Password] Make endpoint available without authentication

### LINKED ISSUE #1449: [Reset Password] Make endpoint available without authentication
Hi Team,

The currently used endpoint to trigger the reset email is protected. 
Please remove the auth check from `/pimcore-studio/api/user/reset-password`

----
Edit:
After the reset password link has been sent, users should be able to set a new password without needing to enter their old one.
Maybe we can pass the Token?

lukmzig:
Resolves #1449 
Needs https://github.com/pimcore/demo-enterprise/pull/443

### DIFF
--- a/src/User/Controller/ResetPasswordController.php
+++ b/src/User/Controller/ResetPasswordController.php
@@ -28,6 +28,7 @@ use Pimcore\Bundle\StudioBackendBundle\Util\Constant\HttpResponseCodes;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Attribute\MapRequestPayload;
 use Symfony\Component\Routing\Attribute\Route;
+use Symfony\Component\Security\Http\Attribute\IsGranted;
 use Symfony\Component\Serializer\SerializerInterface;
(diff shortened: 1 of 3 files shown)


src/User/Controller/ResetPasswordController.php
```diff
@@ -46,6 +47,7 @@ public function __construct(
      * @throws RateLimitException|DomainConfigurationException|SendMailException
      */
     #[Route('/user/reset-password', name: 'pimcore_studio_api_user_reset_password', methods: ['POST'])]
```
martineiber, line 49: We should then also add the Route here into the doc as public access?
https://github.com/pimcore/studio-backend-bundle/blob/1.x/doc/00_Installation.md

martineiber (changes requested):

src/User/Controller/ResetPasswordController.php
lukmzig: Thanks for the hint :)

martineiber (approved):

PR #1540, opened 2025-11-24
[Authentication][Token]:  User is not in the session

lukmzig:
See title

## Additional info
onAuthenticationSuccess() runs before Symfony writes the security token into the session, therefore closing the session here:

- cancels authentication state persistence
- causes user to "lose session after reload"


martineiber (approved):
LGTM

PR #1541, opened 2025-11-24
[Grid] [Filter] Remove all Asset Metadata Filter.

martineiber:
Part of https://github.com/pimcore/studio-ui-bundle/issues/2300

## Additional info
System filters can be used
select => system.select
input => system.string
textarea => system.string
asset => system.integer
document => system.integer
objects => system.integer
date => system.datetime
checkbox => system.boolean


lukmzig (approved):
LGTM

PR #1549, opened 2025-11-25
[Grid] sorting for some specific fields leads to error

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/2587

## Additional info
- sort field needs to respect locale passed from FE

### DIFF
--- a/src/MappedParameter/Filter/SortFilter.php
+++ b/src/MappedParameter/Filter/SortFilter.php
@@ -29,7 +29,7 @@ final readonly class SortFilter
 
     public function getKey(): string
     {
-        if ($this->locale !== null && $this->locale !== '') {
+        if (!empty($this->locale)) {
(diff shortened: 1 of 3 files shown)


src/MappedParameter/Filter/SortFilter.php
```diff
@@ -23,10 +23,20 @@
     public function __construct(
         private string $key = 'id',
         private string $direction = SortDirection::ASC->value,
+        private ?string $locale = null
     ) {
     }
 
     public function getKey(): string
+    {
+        if ($this->locale !== null && $this->locale !== '') {
```
martineiber:
```suggestion
        if (!empty($this->locale)) {
```

martineiber (changes requested):

martineiber (approved):
LGTM

PR #1554, opened 2025-11-26
type fix, add pagination and new data object provider

stunnerparas:
Part of https://github.com/pimcore/product-management/issues/840


src/Gdpr/Schema/GdprSearchOptions.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Symfony\Component\Validator\Constraints\Choice;
+use Symfony\Component\Validator\Constraints\Collection;
+use Symfony\Component\Validator\Constraints\Positive;
+use Symfony\Component\Validator\Constraints\Type;
+
+/**
+ * @internal
+ */
+final readonly class GdprSearchOptions
```
martineiber: It is in the Schema Namespace but it is not a Schema.

martineiber (changes requested):

PR #1557, opened 2025-11-26
[Execution Engine] Batch edit all assets fails with a lot of items 

### LINKED ISSUE #1539: [Execution Engine] Batch edit all assets fails with a lot of items
I tried to batch edit an asset folder with 1200 items. The job ended like this:

Image

When I then restarted it no more messages where executed.

lukmzig:
Resolves #1539 

## Additional info

Batch patch endpoints were adapted so they process each item in a separate instance of a step.
Relevant endpoints were changed as well, they accept only 1 folder per request and the `folder id` parameter was moved to the path. 
Affected endpoints:
- `/pimcore-studio/api/assets/folder` => `/pimcore-studio/api/assets/folder/{id}`
- `/pimcore-studio/api/data-objects/folder` => `/pimcore-studio/api/data-objects/folder/{id}`

Export endpoints were changed similarly as well. However, we had to find a different approach here as the processing of the items has to be done sequentially to preserve the order of the exported elements. Because of this, we had to split the batch export (folder export) for both XLSX and CSV into **2 jobs**.
Important change here is, that the initial job has a `jobRunChildId` which then collects all data in separate steps and creates an export file. For downloading the export file you need to therefore pass **`jobRunChildId`**  and NOT original jobRunId!
Relevant endpoints were changed as well, they accept only 1 folder per request and the `folder id` parameter was moved to the path. 
Affected endpoints:
- `/pimcore-studio/api/export/csv/folder` => `/pimcore-studio/api/export/csv/folder/{id}`
- `/pimcore-studio/api/export/xlsx/folder` => `/pimcore-studio/api/export/xlsx/folder/{id}`

Furthermore, `classId` was added as a required parameter for selected elements exports, when exporting `data-object` elements. This was already the case for `folder exports`

Affected endpoints:
- `/pimcore-studio/api/export/csv`
- `/pimcore-studio/api/export/xlsx`


mcop1 (approved):

lukmzig:
Waiting for FE adjustments and can be merged after

PR #1562, opened 2025-11-27
[Grid] Fix RGBA Filter

martineiber:
Part of https://github.com/pimcore/studio-ui-bundle/issues/2300
## Additional info

alpha is saved as 0 - 255 but the color picker returns 0 - 1


lukmzig (approved):
LGTM

PR #1601, opened 2026-01-15
[PEES-920] Split DDL and DML into different migration.

martineiber:
Resolves #

### DIFF
--- a/src/Migrations/Version20250616120000.php
+++ b/src/Migrations/Version20250616120000.php
@@ -24,7 +24,12 @@ final class Version20250616120000 extends AbstractMigration
 {
     public function getDescription(): string
     {
-        return 'Hide already existing studio job runs';
+        return 'Hide already existing studio job runs (create the schema)';
+    }
+
+    public function isTransactional(): bool
+    {
+        return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/Migrations/Version20250616120000.php
```diff
@@ -24,7 +24,12 @@ final class Version20250616120000 extends AbstractMigration
 {
     public function getDescription(): string
     {
-        return 'Hide already existing studio job runs';
+        return 'Hide already existing studio job runs (create the schema)';
+    }
+
+    public function isTransactional(): bool
```
lukmzig, line 30: So we need this at the end?

lukmzig (approved):

src/Migrations/Version20250616120000.php
martineiber: Yes DDL operations have a implicit commit. So we should not wrap it with a Transaction.
For more infos see the here the Doctrine Docs:
https://www.doctrine-project.org/projects/doctrine-migrations/en/3.9/explanation/implicit-commits.html

lukmzig: thanks for the info :)

PR #1611, opened 2026-01-20
[Grid] List available columns for allowed field of Relation adjustments

### LINKED ISSUE #1389: [Grid] List available columns for allowed field of Relation adjustments
See https://github.com/pimcore/studio-backend-bundle/pull/1193

1. The passed relationField does not support out dot-notation (https://github.com/pimcore/studio-backend-bundle/pull/1084), right? Without the dot-notation it's impossible to use it e.g. within object bricks

2. When I select News and Events:

Image

It's also possible to select the `title` field in the class editor as this is a field which is contained in all classes:

Image

The `List available columns for allowed field of Relation` endpoint does not deliver those columns.

martineiber:
Resolves #1389

### DIFF
--- a/src/Grid/Service/ColumnConfigurationForRelationService.php
+++ b/src/Grid/Service/ColumnConfigurationForRelationService.php
@@ -191,7 +191,7 @@ final readonly class ColumnConfigurationForRelationService implements ColumnConf
             $unique = array_values(array_unique($frontendTypes));
 
             if (count($unique) === 1) {
-                $result[$key] = $configs[0]; // or reset($configs)
+                $result[$key] = $configs[0];
             }
         }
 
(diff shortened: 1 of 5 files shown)


lukmzig (approved):

src/Grid/Service/ColumnConfigurationForRelationService.php
```diff
@@ -155,23 +150,51 @@ private function extractVisibleGridFields(ManyToManyObjectRelation $fieldDefinit
...
+
+            $frontendTypes = array_map(
+                static fn ($c) => $c->getFrontendType(),
+                $configs
+            );
+
+            $unique = array_values(array_unique($frontendTypes));
+
+            if (count($unique) === 1) {
+                $result[$key] = $configs[0]; // or reset($configs)
```
lukmzig: we can remove this comment

PR #1613, opened 2026-01-21
[Documents]: Allow template setting when adding new document

lukmzig:
Part of https://github.com/pimcore/headless-documents/issues/229

## Additional info
When adding a new document, it should be possible to set template in the initial request
This is needed e.g for headless documents

### DIFF
--- a/src/Document/Schema/DocumentAddParameters.php
+++ b/src/Document/Schema/DocumentAddParameters.php
@@ -24,7 +24,7 @@ use Pimcore\Bundle\StudioBackendBundle\Util\Constant\Document\DocumentTypes;
 #[Schema(
     title: 'DocumentAdd',
     required: [
-        'key', 'type', 'title', 'navigationName', 'docTypeId',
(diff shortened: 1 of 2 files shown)


src/Document/Schema/DocumentAddParameters.php
```diff
@@ -24,7 +24,7 @@
 #[Schema(
     title: 'DocumentAdd',
     required: [
-        'key', 'type', 'title', 'navigationName', 'docTypeId',
+        'key', 'type', 'title', 'navigationName', 'docTypeId', 'template',
```
martineiber, line 27: Does not need to be required if the property is nullable?

martineiber (approved):

src/Document/Schema/DocumentAddParameters.php
lukmzig: afaik we make it reqiured for all which are present in schema always

PR #1626, opened 2026-01-23
Include property data as its versioned.

torqdev:
Property data is versioned data, and should be included with the DataObjectVersion data so it can be displayed on the Versions tab. 

This pull requests add the Property schema to the version data. A separate pull request to the Studio-UI bundle will be required to display the property updates / changes on the UI.


martineiber (approved):
LGTM

Thank you for your contribution.

PR #1705, opened 2026-03-16
[Grid] AdvancedColumnCollector exception

### LINKED ISSUE #1652: [Grid] AdvancedColumnCollector exception
Hi Team, 

If you open the data-object folder with the ID 695 you get several endpoint errors. 

See: 
http://localhost/pimcore-studio/api/data-object/grid/available-columns?classId=EF_FD&folderId=695
http://localhost/pimcore-studio/api/data-object/grid/configuration/695/EF_FD

I added the following lines to check if that already fixes the issue:

```php
if($classDefinition !== null) {
    $classIds[] = $classDefinition->getId();
}
```

That brings then the following error: 

Image

Best guess would be that back in the days `Category` was named `ProductCategory` and this name is still in there. 
Please have a look and check why that occures and how to prevent it.

lukmzig:
Resolves #1652

## Additional info
Add check if the class definition exists in AdvancedColumnCollector

### DIFF
--- a/src/Grid/Column/Collector/DataObject/AdvancedColumnCollector.php
+++ b/src/Grid/Column/Collector/DataObject/AdvancedColumnCollector.php
@@ -13,10 +13,8 @@ declare(strict_types=1);
 
 namespace Pimcore\Bundle\StudioBackendBundle\Grid\Column\Collector\DataObject;
 
-use Exception;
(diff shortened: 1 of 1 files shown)


martineiber (approved):
LGTM

Issue #1680, opened 2026-03-02
[GDPR] Permission Issue


Corepex:
Original Issue: https://github.com/pimcore/studio-ui-bundle/issues/2890

If you are logged in as a superuser, you get an error message when you try to search for a user.

<img width="1422" height="133" alt="Image" src="https://github.com/user-attachments/assets/3ff277f1-742e-4094-b312-56f2bde21c92" />

PR #1737, opened 2026-03-31
[GDPR] fix permission issues when exporting elements

### LINKED ISSUE #2886: [GDPR] export should not work for elements without permission
**Steps to reproduce**
- Login as admin
- Create or adapt a user and exclude some data objects
- Login as this user
- Search for those objects in the GDPR extractor
- Export one of the data objects

**Actual behaviour**
- Data object can be exported without permission for this data object

**Expected behaviour**
- It should not be possible to extract data of a data object without permission

**Environment**

- Studio Nightly
- Windows
- Google Chrome

wwidergoldpimcore:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/2886

## Additional info
Check for view permissions when accessing Objects & Assets.
Fix `PimcoreUserProvider` requiring `UserPermissions::PIMCORE_USER` instead of `UserPermissions::USER_MANAGEMENT`.


wwidergoldpimcore:
The Permission in the `PimcoreUserProvider` might also relate to the error described in https://github.com/pimcore/studio-backend-bundle/issues/1680

martineiber (approved):

PR #1749, opened 2026-04-07
[Class definition]: Guess type based on the value of class definitions

lukmzig:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/3247

## Additional info
- As core class definitions supports only string for the values (there are no types) we need to guess the type basedo n this value in our class hydrator


martineiber (approved):
LGTM

PR #1754, opened 2026-04-13
[Select options] - configuration should not be editable for non admins if 'admin only' set to true

### LINKED ISSUE #3162: [Select options] - configuration should not be editable for non admins if 'admin only' set to true
**Steps to reproduce**
- Login as admin
- Create a select option configuration and set 'Admin only' to true
- Login as superuser
- Search for this configuration
- Do a change in the configuration and save

**Actual behaviour**
- Configuration can be changed
- Error message appears when saving

**Expected behaviour**
- Configuration should be read only

**Environment**

- Studio Nightly
- Windows
- Google Chrome

lukmzig:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/3162

## Additional info
consider adminOnly for isWriteable flag


martineiber (approved):
LGTM

PR #1764, opened 2026-04-20
[DataObjects][Calculated Values] getCalculatedValueForEditMode not considered

martineiber:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/3125

## Additional info
Localized Fields
Classificationstore
ObjectBricks

Needs 
https://github.com/pimcore/static-resolver-bundle/pull/137

### DIFF
--- a/src/DataObject/Data/DetailDataInterface.php
+++ b/src/DataObject/Data/DetailDataInterface.php
@@ -1,0 +1,37 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * This source file is available under the terms of the
+ * Pimcore Open Core License (POCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (https://www.pimcore.com)
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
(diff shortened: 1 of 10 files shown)


src/DataObject/Data/DetailDataInterface.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\DataObject\Data;
+
+use Pimcore\Bundle\StudioBackendBundle\DataObject\Data\Model\FieldContextData;
+use Pimcore\Model\DataObject\ClassDefinition\Data;
+use Pimcore\Model\DataObject\Concrete;
+
+/**
+ * @internal
```
lukmzig: Should we keep this internal?

martineiber: Good Point. I will also add a doc part for this.

martineiber: Done in [ccea0c1](https://github.com/pimcore/studio-backend-bundle/pull/1764/commits/ccea0c1f67a9ce4e911c0d9cde948099649962fb)

lukmzig (approved):

PR #1770, opened 2026-04-21
[Task] Add mercure host resolution fallback

### LINKED ISSUE #1501: Studio breaks without a domain in the website settings
The mercure services are using `new Url(...)` to resolve the domain and add the corresponding topics as parameters.
Unfortunately this can cause Errors, when the domain in the website settings are empty. 
Since `new Url(...)` expects a domain, but delivered is only a slug.

martineiber:
Resolves #1501

### DIFF
--- a/src/Mercure/Service/UrlService.php
+++ b/src/Mercure/Service/UrlService.php
@@ -13,8 +13,9 @@ declare(strict_types=1);
 
 namespace Pimcore\Bundle\StudioBackendBundle\Mercure\Service;
 
-use Pimcore\Bundle\StaticResolverBundle\Lib\ToolResolverInterface;
 use Pimcore\Bundle\StudioBackendBundle\Mercure\Util\Constant\Mercure;
+use Symfony\Component\HttpFoundation\RequestStack;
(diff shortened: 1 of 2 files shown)


src/Mercure/Service/UrlService.php
```diff
@@ -43,16 +44,30 @@ public function getClientSideUrl(): string
...
+
+    /**
+     * @throws LogicException
+     */
+    private function getHostUrl(): string
+    {
+        $request = $this->requestStack->getMainRequest();
+
+        if ($request === null) {
+            throw new LogicException('Mercure fallback URL resolution requires an active HTTP request.');
```
lukmzig, line 68: Should we not rather use some studio exception?

martineiber: When it is not a HTTP Request an API Exception has no effect. I think for that case LogicException is fine?

lukmzig: Ah I see ...in this case makes sense, you are right

lukmzig (approved):

PR #1773, opened 2026-04-21
[Custom Reports]: Make element_type and element_class_id optional in CsvCreationHandler

markus-moser:
## Problem

The custom report CSV export fails with:

> The required options "element_class_id", "element_type" are missing.

`CsvCreationHandler` uses `ExportCreationHandlerSetupTrait` which marks `element_type`, `element_class_id`, and `columns` as required step config fields. This is correct for element grid exports (assets, data-objects), but the custom report flow reuses the same `CsvCreationMessage` / `CsvCreationHandler` without those fields — custom reports are not tied to any element type, so the collection step pre-builds all CSV data itself.

## Fix

- Remove `ExportCreationHandlerSetupTrait` from `CsvCreationHandler` and override `configureStep` directly: only `CONFIG_CONFIGURATION` and `CONFIG_COLUMNS` remain required (with empty-array defaults); `ELEMENT_TYPE` and `ELEMENT_CLASS_ID` become optional with `null` defaults.
- Add `extractOptionalConfigFieldFromJobStepConfig` helper to `AbstractHandler` for safe extraction with a default value.
- Switch `CsvCreationHandler` to use the optional extractor for `element_type` and `element_class_id`.

The existing element-grid CSV export flow is unaffected — it still provides all fields, they are just no longer enforced at the step-config validation level.

### DIFF
--- a/src/ExecutionEngine/AutomationAction/AbstractHandler.php
+++ b/src/ExecutionEngine/AutomationAction/AbstractHandler.php
@@ -156,6 +156,16 @@ class AbstractHandler extends AbstractAutomationActionHandler
         throw new EnvironmentException('How did I get here?');
     }
 
+    protected function extractOptionalConfigFieldFromJobStepConfig(
+        GenericExecutionEngineMessageInterface $message,
+        string $key,
+        mixed $default = null,
+    ): mixed {
+        $config = $this->getCurrentJobStepConfig($message);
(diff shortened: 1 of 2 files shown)


martineiber (approved):
LGTM

lukmzig (approved):

PR #1787, opened 2026-04-27
[Data Object Grid] Respect fallback languages for localized fields

markus-moser:
Introduces the `applyFallbackLanguages` flag on the data object grid request body. When set to `true`, empty localized field values walk the configured fallback language chain (and fall back to the system default language as a last resort) instead of returning null.

The flag defaults to `false` so existing grid behaviour is completely unaffected. It is implemented via Pimcore's built-in `Localizedfield::setGetFallbackValues()` mechanism — no custom fallback chain logic.

Companion PR: pimcore/studio-ui-bundle#3424
Part of https://github.com/pimcore/studio-ui-bundle/issues/3405

### DIFF
--- a/src/Grid/Column/Resolver/DataObject/AdapterResolver.php
+++ b/src/Grid/Column/Resolver/DataObject/AdapterResolver.php
@@ -58,11 +58,13 @@ final class AdapterResolver implements
     ) {
     }
 
+    #[\Override]
     protected function doGetFallbackValues(): bool
     {
         return $this->localizedFieldResolver->doGetFallbackValues();
     }
 
+    #[\Override]
     protected function getDefaultLanguage(): ?string
     {
         return $this->toolResolver->getDefaultLanguage();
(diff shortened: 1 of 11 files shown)


src/Grid/Column/Resolver/DataObject/AdapterResolver.php
```diff
@@ -50,11 +52,22 @@ final class AdapterResolver implements
...
 
+    protected function doGetFallbackValues(): bool
+    {
+        return $this->localizedFieldResolver->doGetFallbackValues();
+    }
+
+    protected function getDefaultLanguage(): ?string
+    {
+        return $this->toolResolver->getDefaultLanguage();
+    }
```
martineiber, line 71: Can we add the `#[Override]` Attribute, it is not so easy to see that this overrides something from a Trait.

martineiber (changes requested):

src/Grid/Column/Resolver/DataObject/AdapterResolver.php
markus-moser: Quickly tried it but #Override does not work for traits:
image

I agree that this is not the nicest situation, but don't have a good idea how to solve it better.

markus-moser: Added a phpdoc here:
https://github.com/pimcore/studio-backend-bundle/pull/1787/changes/1ccef8d7aa4cc54c62c89763b037036114bea040

Could you please re-check?

martineiber (approved):
LGTM

PR #1794, opened 2026-04-29
[Website Settings] - stabilize paging ordering of items 

### LINKED ISSUE #1783: [Website Settings] sorting with multiple pages only sorts active page

xIrusux:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/1783

## Additional info
 Now every listing query sorted through this filter will produce e.g. ORDER BY type ASC, id ASC, guaranteeing stable pagination.

### DIFF
--- a/src/Listing/Filter/SortFilter.php
+++ b/src/Listing/Filter/SortFilter.php
@@ -30,8 +30,8 @@ final readonly class SortFilter implements FilterInterface
         }
 
         $sortFilter = $parameters->getSortFilter();
-        $listing->setOrderKey($sortFilter->getKey());
-        $listing->setOrder($sortFilter->getDirection());
+        $listing->setOrderKey([$sortFilter->getKey(), 'id']);
+        $listing->setOrder([$sortFilter->getDirection(), 'asc']);
 
         return $listing;
     }


src/Listing/Filter/SortFilter.php
```diff
@@ -30,8 +30,8 @@ public function apply(
         }
 
         $sortFilter = $parameters->getSortFilter();
-        $listing->setOrderKey($sortFilter->getKey());
-        $listing->setOrder($sortFilter->getDirection());
+        $listing->setOrderKey([$sortFilter->getKey(), 'id']);
```
martineiber, line 33: We cannot guarantee that every listing has an `id` attribute, so this approach will not work.

When i read the issue current. This shoudl already be fixed with https://github.com/pimcore/studio-ui-bundle/pull/3395
Why do we need that?

martineiber: When we have the case that we need sort for multiple fields, we should think about introducing this kind of filter.

xIrusux: @martineiber unfortunately the ordering is not guaranteed even after my last PR. Lets discuss this. :)

Issue #1777, opened 2026-04-22
tags_search permission breaks backward compatibility — existing users get 403 on /api/tags


Maxara:
**Summary**                                                                                                                                                                                                                                                                     
The Studio Backend Bundle introduces a new `tags_search` permission that is required by the `GET /api/tags` endpoint (`CollectionController`). This permission did not exist in the Classic Admin UI, where browsing/searching tags was implicitly allowed for any user with     
`tags_assignment` or `tags_configuration`.                                                                                                                                                                                                                                       
After migrating to Studio UI, existing users who could previously browse tags now receive a 403 Forbidden error because they don't have the new `tags_search` permission.                                                                                                        

**Expected Behavior**                                                                                                                                                                                                                                                            
Users who have `tags_assignment` or `tags_configuration` should be able to browse and search tags in Studio UI without requiring an additional permission — matching the Classic Admin behavior.                                                                                 

**Actual Behavior**                                                                                                                                                                                                                                                               
The `CollectionController::getTags()` endpoint requires `tags_search` via `#[IsGranted(UserPermissions::TAGS_SEARCH->value)]`. Since this permission didn't exist in the Classic Admin, no existing users have it, and they all get 403 when Studio UI tries to load the tag     
tree.

**Root Cause**                                                                                                                                                                                                                                                                    
- **Classic Admin:** Tag browsing only checked `tags_configuration` (in `TagsController`)
- **Studio UI:** Introduced a separate `tags_search` permission but provides no migration to grant it to existing users
- The permission is registered in `users_permission_definitions` but never auto-assigned during the bundle installation or upgrade

**Suggested Fix **                                                                                                                                                                                                                                                                
Add a migration in the Studio Backend Bundle that automatically grants `tags_search` to all users/roles that have `tags_assignment` or `tags_configuration`:                                                                                         

```sql
UPDATE users
SET permissions = CONCAT(permissions, ',tags_search')
WHERE (permissions LIKE '%tags_assignment%' OR permissions LIKE '%tags_configuration%')
AND permissions NOT LIKE '%tags_search%';
```

**Environment**                                                                                                                                                          
- Pimcore Studio Backend Bundle: 2025.4.0
- Pimcore Core: 12.3.x
- Migrating from Classic Admin (pimcore/admin-ui-classic-bundle 2.3.x)

lukmzig:
[...]

PR #1796, opened 2026-04-29
[Tag]: Allow tags_search permission on GET tag endpoints

markus-moser:
Related to #1777

The `GET /tags` and `GET /tags/{id}` endpoints were requiring `tags_configuration`, which blocked users who only have `tags_search` from loading the tag tree in the filter sidebar.

Both endpoints now accept either `tags_configuration` or `tags_search` using the existing `HasOneOfUserPermissionVoter`.


lukmzig (approved):
LGTM

PR #1800, opened 2026-04-30
[Improvement] Performance on batch-delete of objects is quite poor

### LINKED ISSUE #1792: [Improvement] Performance on batch-delete of objects is quite poor
### Studio API version

2025.4

### Steps to reproduce

Delete folder with 1000 objects

### Actual Behavior

takes quite long

### Expected Behavior

let's double check if we can speed this process up a bit

lukmzig:
Resolves #1792 

## Additional info
- add batching to delete job and batch-delete
- remove unnecessary check for hasChildren to reduce DB queries

### DIFF
--- a/src/Element/ExecutionEngine/AutomationAction/Messenger/Handler/BatchDeleteHandler.php
+++ b/src/Element/ExecutionEngine/AutomationAction/Messenger/Handler/BatchDeleteHandler.php
@@ -96,7 +96,7 @@ final class BatchDeleteHandler extends AbstractHandler
                 $jobRun,
                 $stepName,
                 $totalItems,
-                $totalItems,
+                100,
             );
         }
     }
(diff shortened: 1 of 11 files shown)


markus-moser:
@lukmzig I guess the current UI implementation also shows the right progress with the updated approach? Or do you think that I need to adjust it?

lukmzig:
> @lukmzig I guess the current UI implementation also shows the right progress with the updated approach? Or do you think that I need to adjust it?

It should be adjusted - if you delete folder with more than 500 items the progress bar is broken (in some cases its not showing anything or it resets from 100% to 50 % etc.) because we introduce multiple processing steps here based on amount of batches. This is similar to out updated export.

markus-moser:
@lukmzig I adjusted the frontend, see https://github.com/pimcore/studio-ui-bundle/pull/3625. Could you please merge the backend PR if it is fine? Then I will merge the frontend part too...

martineiber (approved):

PR #1801, opened 2026-04-30
[Feature]: Expose jobName on running-jobs API response and add distinct job names for folder operations (#3455)

markus-moser:
## Summary

- Exposes `jobName` on the `JobRun` API response so the frontend can rehydrate running jobs after a page reload
- Adds distinct job names for folder operations (`PATCH_FOLDER_ELEMENTS`, `COLLECT_CSV/XLSX_FOLDER_EXPORT_ELEMENTS`) so the frontend can distinguish them from their single-element counterparts
- Adds migration to hide existing job runs of the new job types on upgrade

## Related

Part of pimcore/studio-ui-bundle#3455
Related to pimcore/studio-ui-bundle#3468

### DIFF
--- a/src/Migrations/Version20260505124233.php
+++ b/src/Migrations/Version20260505124233.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This source file is available under the terms of the
+ * Pimcore Open Core License (POCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (https://www.pimcore.com)
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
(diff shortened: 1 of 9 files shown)


src/Migrations/Version20260505124233.php
```diff
@@ -0,0 +1,42 @@
...
+    public function up(Schema $schema): void
+    {
+        $this->addSql('
+            INSERT INTO ' . JobRunHidden::TABLE_NAME . ' (jobRunId)
+            SELECT jr.id
+            FROM generic_execution_engine_job_run jr
+            LEFT JOIN ' . JobRunHidden::TABLE_NAME . " jrh ON jr.id = jrh.jobRunId
+            WHERE jrh.jobRunId IS NULL
+            AND jr.executionContext IN ('studio_stop_on_error', 'studio_continue_on_error')
+        ");
```
martineiber, line 40: Seams this is the same Migration like https://github.com/pimcore/studio-backend-bundle/blob/cb601012d44f11332fbe38cac5e76981aa888ba8/src/Migrations/Version20260115100546.php
Why do we need it again?

markus-moser: because if someone already executed the other one... we could als rename the other one to a current timestamp, might be better. but we need to ensure that we start with an empty list otherwise a lot of old jobs will pop-up.

martineiber (approved):
LGTM

PR #1818, opened 2026-05-08
[Bug] fix missing select options

### LINKED ISSUE #1806: [Bug]:  Select with Options Source does not have any options available in Data Classes
### Studio API version

2026.1.1

### Steps to reproduce

1. Create Select Options (ex. TestCat)
2. Add Options
3. Save
4. Navigate to Classes
5. Add Data Component - Select
6. Options Source - Select Options

Image

8. Select Options - Select 'TestCat'
9. Navigate to Data Object Tree
10. Add a new data object / Select existing data object
11. Navigate to Select Data Component added in node 5. The option list is empty
 Image

Definition options is null

Image

Log:
```
pimcore.ERROR: Cannot instantiate enum Pimcore\Model\DataObject\SelectOptions\TestCat
```

Identified the Root Cause:
Pimcore 12 will auto populate `Pimcore\\Bundle\\CoreBundle\\OptionsProvider\\SelectOptionsOptionsProvider`  for `Options Provider Class or Service Name:` for enum resolver.  Pimcore 2026 will remains empty, thus no resolver for enum.

### Actual Behavior

The select drop down showing empty option.

### Expected Behavior

The select drop down should have options added from Select Options.

wwidergoldpimcore:
Resolves #1806 

## Additional info
* Add default for the `optionsProviderClass` if `select_options` was chosen for a select field, similar to old frontend logic

### DIFF
--- a/src/Class/Repository/ClassDefinitionRepository.php
+++ b/src/Class/Repository/ClassDefinitionRepository.php
@@ -335,7 +335,6 @@ readonly class ClassDefinitionRepository implements ClassDefinitionRepositoryInt
         return $values;
     }
 
-    // TODO: test nesting & multiselect functionality
     private function applyOptionsProviderDefaults(array $config): array
     {
         $type = $config['optionsProviderType'] ?? null;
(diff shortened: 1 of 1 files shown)


martineiber (approved):
LGTM

PR #1825, opened 2026-05-11
[Classification Store] - integrity constraint violation

lukmzig:
Resolves https://github.com/pimcore/studio-ui-bundle/issues/3527

## Additional info
When value is null we need to clear the container and return it


martineiber (approved):
LGTM

PR #1835, opened 2026-05-18
Fix JobRun schema serializing jobRunChildId as childJobRunId

markus-moser:
The getter was named `getChildJobRunId()` instead of `getJobRunChildId()`. Symfony's serializer derives the JSON key from the getter name, so the field was serialized as `childJobRunId` in the API response instead of `jobRunChildId`. This prevented the frontend from building parent-child job run chains during rehydration, causing parent-child jobs (zip upload, clone, folder export) to show "finished" state while the child job was still running.

Fixes https://github.com/pimcore/studio-ui-bundle/issues/3524


martineiber (approved):
LGTM

PR #1851, opened 2026-05-26
[Object Editor] - add viewports and languages to object editor preview

lukmzig:
Part of https://github.com/pimcore/studio-ui-bundle/issues/3411

## Additional info
- add missing config data to the layout endpoint
- add possibility to add custom parameters to preview generation endpoint

### DIFF
--- a/src/DataObject/Controller/PreviewController.php
+++ b/src/DataObject/Controller/PreviewController.php
@@ -17,14 +17,18 @@ use OpenApi\Attributes\Get;
 use Pimcore\Bundle\StudioBackendBundle\Controller\AbstractApiController;
 use Pimcore\Bundle\StudioBackendBundle\DataObject\MappedParameter\PreviewParameter;
 use Pimcore\Bundle\StudioBackendBundle\DataObject\Service\PreviewUrlServiceInterface;
+use Pimcore\Bundle\StudioBackendBundle\Exception\Api\InvalidArgumentException;
+use Pimcore\Bundle\StudioBackendBundle\Exception\Api\NotFoundException;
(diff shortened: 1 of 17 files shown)


src/DataObject/Controller/PreviewController.php
```diff
@@ -48,15 +55,33 @@ public function __construct(
...
-    #[IntParameter(name: 'site', description: 'Site ID', required: false)]
+    #[IntParameter(
+        name: 'site',
+        description: 'Site ID for multi-site setups',
+        required: false,
+        example: 0,
+    )]
+    #[TextFieldParameter(
+        name: 'locale',
+        description: 'Any additional query parameter is forwarded to the preview generator.',
```
martineiber: Name and description do not add up somehow.

martineiber (changes requested):

src/DataObject/Controller/PreviewController.php
lukmzig: "locale" name is an example of additional query parameter as an hint. Any parameter can be forwarded to the preview generator. Any better idea how to document this in swagger?

martineiber: What if we remove it completely. And add it to the Description of the endpoint?

lukmzig: We can also do that, yes. I will change it

lukmzig: updated in https://github.com/pimcore/studio-backend-bundle/pull/1851/commits/82bf7df850b5cfda128e7b67424d00fc280bc0b6

martineiber (approved):
LGTM

PR #1852, opened 2026-05-26
[Bug]: [Documents] Modification metadata mapping

### LINKED ISSUE #1828: [Bug]: [Documents] modificationDate remains unchanged on document update
### Studio API version

2025.4

### Steps to reproduce

**Endpoint**: /pimcore-studio/api/documents/{id}, PUT method
**Payload**: [payload.json](https://github.com/user-attachments/files/27628880/payload.json)

Image

### Actual Behavior

After updating a document via a PUT request, the **modificationDate** field remains unchanged. The document content is updated successfully, but the timestamp still reflects the initial value. I noticed that **userModification** field contains the updated date, but **modificationDate** should be updated as well.

On the frontend side, we use the **modificationDate** field as a unified field for all element types.

### Expected Behavior

When updating a document, the **modificationDate** should be updated to the latest modification date after a successful update.

wwidergoldpimcore:
Resolves #1828

### DIFF
--- a/src/Document/Service/HydratorService.php
+++ b/src/Document/Service/HydratorService.php
@@ -58,7 +58,7 @@ final readonly class HydratorService implements HydratorServiceInterface
             $item->getPath(),
             $this->iconService->getIconForDocument($item->getType()),
             $item->getUserOwner(),
-            $item->getModificationDate(),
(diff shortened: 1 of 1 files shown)


martineiber (approved):
LGTM

PR #1872, opened 2026-06-08
[BUG] Gate open and remove actions in relation tables on element permissions

Jonathon-Meney-Torq:
## Summary

- Added `hasAccess` and `canEdit` boolean flags to `RelatedElementData` schema
- `ElementDataService::getRelatedElementData()` computes per-element access using classic admin `AbstractElement::isAllowed()` — `hasAccess` from `view` permission on the related element, `canEdit` from `save` or `publish` permission on the parent object
- Introduced `RelationNormalizationContext` service to pass the parent `Concrete` object into `ElementDataService` during normalization without changing interface signatures
- `DataService::getDetailObjectData()` sets and clears the context via try/finally
- `GridSearch::searchElementsForUser()` skips workspace path and user restrictions when the request is a `system.ids` filter query, so custom field columns load for relation items regardless of workspace access
- `ElementProviderTrait::getElementType()` visibility changed from `private` to `protected` to allow subclassing

### DIFF
--- a/src/DataIndex/Grid/GridSearch.php
+++ b/src/DataIndex/Grid/GridSearch.php
@@ -113,8 +113,21 @@ final readonly class GridSearch implements GridSearchInterface
         UserInterface $user
     ): AssetSearchResult|DataObjectSearchResult|DocumentSearchResult {
         $type = $this->getStudioElementType($type);
-        /** @var AssetQueryInterface|DataObjectQueryInterface|DocumentQueryInterface $query */
-        $query = $this->getSearchQuery($type, $gridParameter, $user);
+
+        // When filtering by explicit IDs (relation context), skip workspace and path
+        // restrictions — the user already has a reference to these objects via the
+        // relation, so showing their field data does not expose new information.
+        $filter = $gridParameter->getFilters();
+        $byIds = $filter->getSimpleColumnFilterByType('system.ids') !== null;
+
+        if ($byIds) {
+            $query = $this->queryFactory->create($type);
+            /** @var AssetQueryInterface|DataObjectQueryInterface|DocumentQueryInterface $query */
+            $query = $this->filterService->applyFilters($query, $filter, $type);
+        } else {
+            /** @var AssetQueryInterface|DataObjectQueryInterface|DocumentQueryInterface $query */
+            $query = $this->getSearchQuery($type, $gridParameter, $user);
+        }
 
         return match($type) {
             ElementTypes::TYPE_ASSET => $this->assetSearchService->searchAssets($query),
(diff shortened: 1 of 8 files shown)


src/Element/Service/ElementDataService.php
```diff
@@ -13,31 +13,58 @@
...
 use Pimcore\Model\DataObject\Concrete;
 use Pimcore\Model\Document;
 use Pimcore\Model\Element\ElementInterface;
+use Pimcore\Model\User;
 
 /**
  * @internal
  */
-final readonly class ElementDataService implements ElementDataServiceInterface
+readonly class ElementDataService implements ElementDataServiceInterface
```
martineiber, line 29: Why removing the final?

src/DataIndex/Grid/GridSearch.php
```diff
@@ -113,8 +113,21 @@ public function searchElementsForUser(
...
+        // restrictions — the user already has a reference to these objects via the
+        // relation, so showing their field data does not expose new information.
+        $filter = $gridParameter->getFilters();
+        $byIds = $filter->getSimpleColumnFilterByType('system.ids') !== null;
+
+        if ($byIds) {
+            $query = $this->queryFactory->create($type);
+            /** @var AssetQueryInterface|DataObjectQueryInterface|DocumentQueryInterface $query */
+            $query = $this->filterService->applyFilters($query, $filter, $type);
+        } else {
```
martineiber, line 127: Can you also fix this Codacy Review. We try to avoid else statements.

src/DataIndex/Grid/GridSearch.php
```diff
@@ -113,8 +113,21 @@ public function searchElementsForUser(
...
+        // When filtering by explicit IDs (relation context), skip workspace and path
+        // restrictions — the user already has a reference to these objects via the
+        // relation, so showing their field data does not expose new information.
+        $filter = $gridParameter->getFilters();
+        $byIds = $filter->getSimpleColumnFilterByType('system.ids') !== null;
+
+        if ($byIds) {
+            $query = $this->queryFactory->create($type);
+            /** @var AssetQueryInterface|DataObjectQueryInterface|DocumentQueryInterface $query */
+            $query = $this->filterService->applyFilters($query, $filter, $type);
```
martineiber, line 126: What is the benefit? `$this->getSearchQuery($type, $gridParameter, $user);`should already apply all given filters. Having here a dependency to the `system.ids` here seams for me the wrong place.

martineiber (changes requested):
Can you check the sonar Report for our code style.

martineiber:
Closed due to inactivity

PR #1873, opened 2026-06-08
[DataObject] Resolve nested field definitions for dynamic select options

markus-moser:

The `/data-objects/select-options` endpoint resolved the field definition via `$object->getClass()->getFieldDefinition($fieldName)`, which only finds **top-level** fields. As a result, dynamic select/multiselect options providers on fields nested in **localized fields, blocks, field collections or object bricks** could not be resolved (the endpoint returned not-found).

This change resolves the field definition through the existing `DotNotationParser` (the same parser `PathFormatterService` uses), so nested dot-notation field names resolve correctly. The options-provider context is additionally enriched with the resolved container context (`containerType`, leaf `fieldname`, `subContainerType`, `subContainerKey`), mirroring `PathFormatterService::convert()`.

Companion to the studio-ui-bundle change for dynamic select options providers ([#1034](https://github.com/pimcore/studio-ui-bundle/issues/1034)). Top-level dynamic selects already worked; this adds nested-field support.


martineiber (approved):
LGTM

PR #1935, opened 2026-07-05
fix: prevent invalid getClassDefinitionIcon() method call

sandeeep-prajapati:
Summary

This change prevents calling "getClassDefinitionIcon()" on objects that are not instances of "DataObjectSearchResultItem".

Changes

- Added an "instanceof DataObjectSearchResultItem" check before calling "getClassDefinitionIcon()".
- Prevents potential "Call to undefined method" errors when a "DataObject" instance is passed to "getClassIcon()".

This change is limited to adding the necessary type guard and does not alter the existing behavior for valid "DataObjectSearchResultItem" instances.

### DIFF
--- a/src/Icon/Service/IconService.php
+++ b/src/Icon/Service/IconService.php
@@ -159,7 +159,7 @@ final readonly class IconService implements IconServiceInterface
             }
         }
 
-        if ($dataObject->getClassDefinitionIcon() !== null) {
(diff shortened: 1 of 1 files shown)


martineiber (changes requested):
Seams like a bugfix. Can you Reabase the base branch to 2025.4 then we can include it into the last Bugfix Release.

PR #1945, opened 2026-07-09
[Notes] Fix "Between" date filter and user search

### LINKED ISSUE #1923: [Notes] "Between" Date Filter operator doesn't work
**Endpoint**: /pimcore-studio/api/notes

When the "Between" operator is selected, the list of notes is empty. However, all other operators work as expected.

Image
Image

### LINKED ISSUE #1925: [Notes] Searching for user does not work
**Endpoint**: /pimcore-studio/api/notes

1) When filtering by user name, the backend returns a 500 error. Initially, we get the **userName** field from the list, but the backend expects the **user** field instead. Shouldn't we align them?

Image

2) Even when filtering by the **user** field name, the result is an empty list.
Image

wwidergoldpimcore:
Resolves #1923
Resolves #1925

Both bugs live in `Note/Service/FilterService::applyFieldFilters`, which translates note field filters into SQL:

- **#1923 — "Between" date filter returns empty.** The bind parameter name was derived from the field name, so two conditions on the same field collided. The UI decomposes a date "Between" range into two conditions (`gt` from + `lt` to) on `date`; the second `:date` binding overwrote the first, collapsing the query to `date > to AND date < to` → always empty. Bind parameters are now indexed per filter (`:filter_0`, `:filter_1`, `:minTime_0`/`:maxTime_0`), so the range resolves to `date > from AND date < to`.

- **#1925 — user search fails.** The user column is exposed as `userName` in the `Note` schema and the UI filters on it, but the backend only special-cased `user`, so `userName` generated a condition against a non-existent column → **500**. The `user` branch also lacked a `continue`, emitting a spurious generic column condition on top of the subquery. The `userName` field is now resolved via a subquery against `users`.`name` and short-circuits the generic condition.

## Additional info
- Aligned the backend to the field name the Studio UI actually sends (`userName`, matching the schema property) — **no UI change required**. The previously-handled `user` field name is dropped (it was broken and unused by the UI).
- Regression tests added/updated in `FilterServiceTest` for the "Between" range and the `userName` search; verified they fail without the fix. Full unit suite passes (444 tests), PHPStan clean.

### DIFF
--- a/src/Note/Service/FilterService.php
+++ b/src/Note/Service/FilterService.php
@@ -52,7 +52,7 @@ final readonly class FilterService implements FilterServiceInterface
 
             $propertyKey = 'field';
 
-            foreach ($parameters->getFieldFiltersArray() as $filter) {
+            foreach ($parameters->getFieldFiltersArray() as $index => $filter) {
                 $operator = $this->findOperator($filter['type'], $filter['operator']);
                 $value = $this->prepareValue($filter['type'], $filter['operator'], $filter['value']);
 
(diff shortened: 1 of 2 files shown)


martineiber (approved):
LGTM

PR #1946, opened 2026-07-09
[Listing] Fix date filter "Between" and Unix-timestamp columns

### LINKED ISSUE #1924: [Notifications] "Between" Date Filter operator doesn't work
**Endpoint**: /pimcore-studio/api/notifications

When the "Between" operator is selected, the list of notifications is empty. However, all other operators work as expected.

Image
Image

### LINKED ISSUE #1937: [Recycle Bin] Date Filter operators don't work
**Endpoint**: /pimcore-studio/api/recycle-bin/items

**_The list of elements:_**
Image

1) **On** operator
Image

2) **Between** operator
Image

Please check and fix all the operators for the data filter.

wwidergoldpimcore:
Resolves #1924
Resolves #1937

Both bugs live in the shared `Listing/Filter/DateFilter`, used by the notifications, recycle-bin (and other) listings.

- **#1924 — Notifications "Between" returns empty.** The bind parameter name was derived from the column key, so two conditions on the same column collided. The UI sends a "Between" range as two `columnFilters` (`from` + `to`) on the same key (`creationDate`); both bound to `:creationDate`, the second overwrote the first, and the query collapsed to `creationDate >= to AND creationDate <= to` → always empty. This is the same class as #1923. Bind parameters are now indexed per applied filter (`:filter_0`/`:filter_1`, `:minTime_0`/`:maxTime_0`).

- **#1937 — Recycle bin date filters don't work (all operators).** In addition to the collision above, the filter always compared against a Carbon **datetime string** (`'2024-06-10 00:00:00'`), but `recyclebin.date` is an `int(11) unsigned` **Unix timestamp**. MySQL coerced the string to a number, so even "On" returned an empty list. Columns that store a Unix timestamp are now compared against an integer timestamp.

### Why a class-keyed map for the timestamp columns
The generic `DateFilter` has no reliable way to learn a column's storage type: the frontend sends a generic `date` type, listings expose no table-name accessor (the recycle-bin Dao hardcodes `FROM recyclebin` in raw SQL), and per-request schema introspection would be heavy and fragile. The storage format is a fixed, domain-specific fact, so it's encoded explicitly as `UNIX_TIMESTAMP_COLUMNS` (listing class → column keys). Notifications (`creationDate` is a `TIMESTAMP`) keeps the datetime-string path and is unaffected.

## Additional info
- Regression tests added to `DateFilterTest` for the "Between" collision and the Unix-timestamp column; verified they fail without the fix. Full unit suite passes (446 tests), PHPStan clean.
- **Follow-up (studio-ui-bundle):** the recycle-bin date filter column is currently commented out with a `TODO` referencing #1937 (`recycle-bin/filters/filters.tsx`). It should be re-enabled once this lands.
- Separate from #1945 (Notes `FilterService`), which is a different filter implementation.

### DIFF
--- a/src/Listing/Filter/DateFilter.php
+++ b/src/Listing/Filter/DateFilter.php
@@ -18,6 +18,7 @@ use Pimcore\Bundle\StaticResolverBundle\Db\DbResolverInterface;
 use Pimcore\Bundle\StudioBackendBundle\Exception\Api\InvalidArgumentException;
 use Pimcore\Bundle\StudioBackendBundle\Filter\FilterType;
 use Pimcore\Bundle\StudioBackendBundle\MappedParameter\Filter\ColumnFilter;
+use Pimcore\Model\Element\Recyclebin\Item\Listing as RecycleBinListing;
 use Pimcore\Model\Listing\AbstractListing;
 use function is_array;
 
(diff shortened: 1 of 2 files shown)


martineiber (approved):
LGTM

PR #1953, opened 2026-07-16
[Asset Preview] Persist folder preview image size per user (#3745)

### LINKED ISSUE #3745: [Asset Preview] - save image size settings
Would it be possible to save the image size settings per folder to have the same seetings after a logoff/login?
Now the settings are stored within a session and get lost after logoff/login.

Image

xIrusux:
## What
Backend for pimcore/studio-ui-bundle#3745 — persist the asset folder **Preview** tab image-display size (small/large) **per user + folder**, so it survives logout/login instead of only living in the browser session.

## Changes
- New Doctrine entity `Entity/Asset/FolderPreviewSetting` (composite PK `user` + `assetFolderId`, `imageSize`), created in `Installer` + migration `Version20260715120000` (raw `CREATE TABLE` to avoid the MariaDB `bit`-type schema-diff error).
- `Asset/Repository/FolderPreviewSettingRepository` (get by user+folder, upsert).
- `Asset/Service/FolderPreviewSettingService` — returns default `small` when unset, upserts on save; dispatches a pre-response event.
- `Asset/Schema/FolderPreviewSetting` DTO + `Asset/Hydrator/...` + `Asset/Event/PreResponse/FolderPreviewSettingEvent`.
- Two one-action controllers under `Asset/Controller/FolderPreviewSetting/`: `GET`/`PUT /assets/{folderId}/preview-setting`, guarded with `#[IsGranted(ASSETS)]`, fully OpenAPI-documented (+6 translation keys).

## Tests
- Codeception unit tests for repository, hydrator and service.

## Companion PR
- Frontend: pimcore/studio-ui-bundle (branch `feature/3745-persist-asset-preview-image-size-2026x`).

## Notes
- Follows the layered Controller→Service→Repository→Hydrator/DTO/Event architecture.

### DIFF
--- a/config/assets.yaml
+++ b/config/assets.yaml
@@ -19,6 +19,10 @@ services:
   Pimcore\Bundle\StudioBackendBundle\Asset\Encoder\TextEncoderInterface:
       class: Pimcore\Bundle\StudioBackendBundle\Asset\Encoder\TextEncoder
 
+  # Repositories
+  Pimcore\Bundle\StudioBackendBundle\Asset\Repository\FolderPreviewSettingRepositoryInterface:
+    class: Pimcore\Bundle\StudioBackendBundle\Asset\Repository\FolderPreviewSettingRepository
+
   # Services
   Pimcore\Bundle\StudioBackendBundle\Asset\Service\AssetServiceInterface:
     class: Pimcore\Bundle\StudioBackendBundle\Asset\Service\AssetService
--- a/config/assets.yaml
+++ b/config/assets.yaml
@@ -15,6 +15,9 @@ services:
   Pimcore\Bundle\StudioBackendBundle\Asset\Hydrator\CustomSettingsHydratorInterface:
     class: Pimcore\Bundle\StudioBackendBundle\Asset\Hydrator\CustomSettingsHydrator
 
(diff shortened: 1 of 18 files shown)


martineiber:
TBH, I’m not sure whether creating a separate database entity just for this setting is the best approach.

Could we instead extend the user entity with something like uiSettings? This would give us more flexibility and allow us to store additional UI-related settings in the future.

fashxp:
we also could consider storimg it in local storage client side - similar to the opened tabs? wdyt?

martineiber:
Agreed — unless we need it to sync across browsers/machines, local storage (like we do for the opened tabs) is the cleanest option here. The database approach would be a bit overkill then.

xIrusux:
sounds good! Thanks for the input everyone. Working on it.

xIrusux:
Reimplemented client-side in pimcore/studio-ui-bundle#3897: the size is stored in local storage as a `folderId -> size` map under a user-namespaced key (`pimcore_studio_asset_folder_preview_image_size_<userId>`), reusing the same storage utils and key convention as the persisted widget-manager tab layout. Only folders deviating from the default are kept, and the key is removed once the map is empty.

That means **nothing here is needed**: no entity, no `bundle_studio_asset_folder_preview_settings` table, no migration, no endpoints. Since none of this was ever merged to `2026.x`, no cleanup migration is required either.

Closing this PR.

PR #1960, opened 2026-07-21
[Relation] Tolerate missing column type/config in GridService instead of failing the whole grid

klemensusec:
GridService::getConfigurationFromArray() built each Column with type: $column['type'] / config: $column['config'] and no null-coalescing, so a single column missing type/config aborted the entire grid request with 422 "Invalid column configuration" (and an "Undefined array key type" warning). This broke relation preview grids whose default fallback columns arrive without a type.

Harden the column construction:

- default config to [] when missing
- resolve a missing type for known system columns (id -> system.id, fullpath/classname -> system.string) via a new resolveSystemColumnType()
- skip a single column whose type is still unresolvable, logging a warning, instead of failing the whole request (mirrors the existing index-out-of-sync skip pattern)
- keep throwing InvalidArgumentException only for genuinely malformed input
    (missing key)

Add GridServiceTest coverage: typeless system columns resolve without a 422, and an unresolvable column is skipped rather than fatal.


klemensusec:
I have read the CLA Document and I hereby sign the CLA

PR #1963, opened 2026-07-23
Allow disabling auto-create of missing translation keys

### LINKED ISSUE #2730: Allow disabling auto-create translation keys for studio translation domain
This behavior already existed in Classic UI, but I think it would be a great opportunity to re-think in Studio:
Pimcore studio auto-creates all non-existent translation keys in the translation database. However, I consider it best practice to provide "backend" translations using .yaml files (like /translations/studio.en.yaml for example). Doing so ensures that we have the same translations in dev as well as in staging and production systems. 
If a key is missing in our setups, there's a good reason for it: Usually this happens for numeric or other non-translatable labels mostly from select field choices. In our projects, the "studio" (or formerly the "admin") namespace is polluted with hundreds of translation keys that never need to be translated.
Would it be a possible to disable auto-creating non-existing translations keys in Pimcore studio? This could be a global configuration IMHO, since it still would allow the user to (manually) add translation keys if there are missing some.

ValeriaMaltseva:
## What

Adds a global config flag to disable the automatic creation of missing translation keys for the `studio` translation domain.

```yaml
# config/config.yaml
pimcore_studio_backend:
    translations:
        auto_create_missing_keys: false # default: true
```

The flag is exposed to the Studio frontend via `SystemSettingsProvider` as `auto_create_translations`.

## Why

Resolves [#2730](https://github.com/pimcore/studio-ui-bundle/issues/2730). Projects that maintain the `studio` domain exclusively through YAML files (e.g. `translations/studio.en.yaml`) currently get the `translations_studio` table polluted with hundreds of non-translatable keys (numeric values, select-field choices) that the frontend auto-creates on render. This flag lets them opt out. Default `true` preserves current behaviour.

## Scope / follow-up

This PR is the **backend half**. The flag is currently dormant: the frontend gate that actually stops the auto-create request based on `auto_create_translations` ships **separately in `studio-ui-bundle`** (Path 1 — the i18next `saveMissing` → `translationCreate` flow). Until that lands, behaviour is unchanged.

Manual creation via the Translations editor is unaffected.

> Note: this only covers the frontend-driven auto-creation. Keys the backend itself resolves through the Symfony translator may still be created by Pimcore core's write-on-miss behaviour for registered domains (documented in the Translations doc).

## Docs

Added an "Auto-Creating Missing Keys" section to `doc/03_Extending/07_Translations.md`.

### DIFF
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -51,6 +51,8 @@ services:
     tags: [ 'pimcore.studio_backend.settings_provider' ]
 
   Pimcore\Bundle\StudioBackendBundle\Setting\Provider\SystemSettingsProvider:
+    arguments:
+      $autoCreateTranslations: '%pimcore_studio_backend.translations.auto_create_missing_keys%'
     tags: [
       'pimcore.studio_backend.settings_provider',
       'pimcore.studio_backend.update_settings_provider'
--- a/doc/03_Extending/07_Translations.md
+++ b/doc/03_Extending/07_Translations.md
@@ -38,6 +38,36 @@ This keeps the full catalogue private until the user has logged in.
 
 ---
 
+## Auto-Creating Missing Keys
+
+When the Studio frontend renders a translation key that does not yet exist in the `studio`
+catalogue, it reports the key to the backend, which creates an empty entry in the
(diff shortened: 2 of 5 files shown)


martineiber (approved):
LGTM

PR #1964, opened 2026-07-24
[Permissions] User can save object and asset data even if not allowed to (PEES-1371)

ValeriaMaltseva:
## Issue

Tracking: **[PEES-1371](https://github.com/pimcore/service-operations/issues/1010)** — *[Permissions][Studio UI] User can save any object data even if they are not allowed to.*

## Problem

The element update and patch endpoints only enforced the **global menu permission**
(`#[IsGranted(objects/assets)]`). None of them checked the **per-object workspace
permission** on the write path, so a user who could reach the Objects/Assets area but
lacked edit rights on a specific element could still persist changes:

- `PUT /data-objects/{id}` and `PUT /assets/{id}` (normal edit + auto-save)
- `PATCH /data-objects` and `PATCH /assets` (**Batch Edit**, single and multi-element)

Only `publish`/`unpublish` were guarded (in `ElementSaveService`); the `save`, `autoSave`,
`version` and untasked writes reached `$element->save()` with no workspace check. The
multi-element Batch Edit path did not even check `view`. The workspace flags were correctly
computed and sent to the UI, but never enforced server-side.

## Fix

Enforce the workspace permission at the two write choke points, before any data is persisted:

- **`UpdateService::update()`** and **`PatchService::patchElement()`**
- **Data objects (`Concrete`)** → require the **`save`** permission for `save` / `autoSave` /
  `version` / untasked writes. `publish` / `unpublish` remain gated by their existing checks
  in `ElementSaveService`, so a publish-only user can still Save & Publish.
- **Assets** → require the **`publish`** permission (assets have no `save` workspace
  permission; `publish` is their edit/write permission, matching the Studio UI save button).

A denied write now throws `ForbiddenException` (HTTP 403) regardless of the entry point —
editor save, auto-save, or Batch Edit (including the async multi-element job, which is checked
against the initiating user). Object/asset **creation** paths are untouched (they do not go
through these services).

A new `ElementPermissions::SAVE_PERMISSION` constant is added (there was none).

## Scope note

This PR covers **data objects and assets**. Documents share the same `UpdateService` /
`PatchService` code path and have the same gap; they are intentionally left out of this change
and can be addressed as a follow-up.

## Verification

- `php -l` passes on all changed files.
- The full Codeception suite, php-cs-fixer and PHPStan were **not** run locally (dev
  dependencies are not installed in this checkout) — please rely on CI. No regression test is
  included in this PR yet; a unit test asserting `ForbiddenException` on the object/asset
  write paths for a user lacking the permission should follow.

### DIFF
--- a/src/Patcher/Service/PatchService.php
+++ b/src/Patcher/Service/PatchService.php
@@ -35,11 +35,14 @@ use Pimcore\Bundle\StudioBackendBundle\ExecutionEngine\Util\Config;
 use Pimcore\Bundle\StudioBackendBundle\ExecutionEngine\Util\Jobs;
 use Pimcore\Bundle\StudioBackendBundle\ExecutionEngine\Util\StepConfig;
 use Pimcore\Bundle\StudioBackendBundle\MappedParameter\PatchFolderParameter;
+use Pimcore\Bundle\StudioBackendBundle\Security\Service\SecurityServiceInterface;
 use Pimcore\Bundle\StudioBackendBundle\Updater\Service\UpdateServiceInterface;
 use Pimcore\Bundle\StudioBackendBundle\Util\Constant\DataObject\FieldKeys;
+use Pimcore\Bundle\StudioBackendBundle\Util\Constant\ElementPermissions;
 use Pimcore\Bundle\StudioBackendBundle\Util\Constant\ElementTypes;
 use Pimcore\Bundle\StudioBackendBundle\Util\Constant\PatchDataKeys;
 use Pimcore\Bundle\StudioBackendBundle\Util\Constant\PatcherActions;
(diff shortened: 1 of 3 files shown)


pimcore-deployments:
🚫 **Issue-link guardrail failed — this PR has been converted to draft.**

Every PR must reference a tracking issue in `pimcore/platform-version` so the change stays traceable.

- No issue in `pimcore/platform-version` is linked.

Add the link with a keyword in the **PR description** (not in a comment): https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

When fixed, press **Ready for review** to re-run the checks.

lukmzig (approved):

PR #1981, opened 2026-08-06
[Symfony 8] Inject RateLimiterFactoryInterface instead of the concrete factory

herbertroth:
## What

Replaces the concrete `Symfony\Component\RateLimiter\RateLimiterFactory` with
`RateLimiterFactoryInterface` — **7 site(s) across 6 file(s)** — plus the matching import.

- `doc/02_Installation_and_Configuration/05_Upgrade.md`
- `src/EventSubscriber/RateLimitSubscriber.php`
- `src/Security/RateLimiter/McpLoginRateLimiter.php`
- `src/Setting/Admin/Controller/ThumbnailController.php`
- `src/User/Controller/ResetPasswordController.php`
- `src/User/RateLimiter/RateLimiter.php`
- `src/User/RateLimiter/RateLimiterInterface.php`

Only three kinds of edit: the import, the type-hints, and (where present) a return type. No behaviour,
service definition or test changes.

## Why

Symfony 8.0 **removes the autowiring aliases** for the concrete `RateLimiterFactory`
(`UPGRADE-8.0.md`, FrameworkBundle and SecurityBundle; deprecated since 7.3).

**Behaviour is unchanged on 7.4**, verified against the installed tree (`symfony/rate-limiter v7.4.14`):
`RateLimiterFactoryInterface` has existed **since 7.3**, `final class RateLimiterFactory implements
RateLimiterFactoryInterface`, and its only method — `create(?string $key = null): LimiterInterface` — is
signature-identical. This code calls nothing but `create()`.

Worth being precise about the urgency: the injections here are bound explicitly to a `@limiter.*` service,
whose class stays `RateLimiterFactory` in 8.0, so this particular site would not break at the bump. The
swap is hygiene and future-proofing — a limiter reconfigured as a **compound** limiter is a
`CompoundRateLimiterFactory`, which does *not* extend the concrete class and would only satisfy the
interface type.

Neither PHPStan nor runtime deprecations flag this — the deprecation fires on the alias, and only for
autowired injections. It was found by grep during the platform analysis.

## ⚠️ This one touches public API (please carry into the release notes)

`data-hub-simple-rest` and `openid-connect` got the identical change on `@internal final` classes. Here,
one of the six files is public API:

`Pimcore\Bundle\StudioBackendBundle\User\RateLimiter\RateLimiterInterface` has **no `@internal` marker**,
and `config/users.yaml` maps it as a swappable service — so custom implementations are a supported
scenario. Its only method exposes the concrete factory directly:

```diff
-public function check(RateLimiterFactory $rateLimiterFactory): void;
+public function check(RateLimiterFactoryInterface $rateLimiterFactory): void;
```

**Impact — implementers, not callers:**

- **Callers keep working, unchanged.** `RateLimiterFactory` implements `RateLimiterFactoryInterface`, so
  every existing call site still type-checks. Verified: in-repo `ResetPasswordController` and
  `ThumbnailController`, and cross-repo `pimcore/openid-connect`'s `EndpointController` / `ScriptController`,
  which import this interface and call `check()`.
- **Third-party implementations break** — a narrower parameter type violates contravariance and PHP raises
  a fatal error. They must widen the parameter to `RateLimiterFactoryInterface` too.

**Why widen now rather than deprecate:** Symfony 8.0 does not force this particular signature — the class
survives, only its autowiring alias goes. But PHP cannot accept both signatures at once, so a deprecation
cycle cannot soften the change; it only delays it. Doing it in 2026.3 means integrators adjust once, on a
release that documents it, instead of during the Symfony 8 bump. Documented in
`doc/02_Installation_and_Configuration/05_Upgrade.md` (added in this PR).

A related follow-up, not done here: the analysis suggests marking this interface `@internal` in the next
major if custom rate limiters were never an intended extension point.
## Verification

The central check gained two rules for this in
pimcore/workflows-collection-public#156, merged **before** this PR on purpose, so
**`api-scan (ratelimiter-factory-typehint)` and `api-scan (ratelimiter-factory-import)` on this PR are
the test** — both were failing on this branch's base and must be green here.

Locally the gates ran all five central rules extracted from the merged workflow (clean after the edit),
plus `php -l`, a php-cs-fixer `ordered_imports` check, and a diff-shape check.

## Context

Sweep 3 of the Symfony 7.4 → 8 migration (`pimcore/platform-version`, `migration-analysis/SUMMARY.md` §3,
runbook `EXEC-RATELIMITER.md`): 11 sites across 9 files in 3 repos, one PR each. Sweeps 1 and 2
(`#[TaggedIterator]` → `#[AutowireIterator]`, and the voter `?Vote` argument) are merged.

### DIFF
--- a/doc/02_Installation_and_Configuration/05_Upgrade.md
+++ b/doc/02_Installation_and_Configuration/05_Upgrade.md
@@ -2,6 +2,37 @@
 
 The following steps are necessary during updating to newer versions.
 
+## Upgrade to 2026.3.0
+
+### `RateLimiterInterface::check()` now expects `RateLimiterFactoryInterface`
+
+Symfony 8.0 removes the autowiring aliases for the concrete
+`Symfony\Component\RateLimiter\RateLimiterFactory`. All injections of that class in this bundle were
+changed to `Symfony\Component\RateLimiter\RateLimiterFactoryInterface`, including the public
+`Pimcore\Bundle\StudioBackendBundle\User\RateLimiter\RateLimiterInterface`:
+
+```diff
+-public function check(RateLimiterFactory $rateLimiterFactory): void;
++public function check(RateLimiterFactoryInterface $rateLimiterFactory): void;
+```
+
+**[Code BC break] If you implement `RateLimiterInterface` yourself, you have to widen the parameter type
+as well** — keeping the concrete `RateLimiterFactory` violates parameter contravariance and PHP raises a
+fatal error (`Declaration of ... must be compatible with ...`). The service is swappable via
+`config/users.yaml`, so custom implementations are a supported scenario.
+
+```php
+use Symfony\Component\RateLimiter\RateLimiterFactoryInterface;
+
+public function check(RateLimiterFactoryInterface $rateLimiterFactory): void
+```
+
+**Callers are not affected.** `RateLimiterFactory` implements `RateLimiterFactoryInterface`, so any code
+passing a limiter factory into `check()` keeps working unchanged.
+
+Behaviour is unchanged otherwise: `RateLimiterFactoryInterface` exists since Symfony 7.3 and its only
+method, `create()`, is signature-identical to the concrete class's.
+
 ## Upgrade to 2025.4.6
 - [GDPR Objects Export] Fixed: the data object export now includes inherited values (including inherited localized fields).
 
--- a/src/EventSubscriber/RateLimitSubscriber.php
+++ b/src/EventSubscriber/RateLimitSubscriber.php
@@ -20,7 +20,7 @@ use Symfony\Component\HttpKernel\Event\RequestEvent;
 use Symfony\Component\HttpKernel\Event\ResponseEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\RateLimiter\RateLimit;
-use Symfony\Component\RateLimiter\RateLimiterFactory;
+use Symfony\Component\RateLimiter\RateLimiterFactoryInterface;
 
 /**
  * @internal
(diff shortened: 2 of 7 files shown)


herbertroth:
Closing this — the sweep is being backed out, and the two central `ratelimiter-factory-*` rules are being
removed from `reusable-symfony8-compat.yaml` along with it.

**Why:** widening the *public* `StudioBackendBundle\User\RateLimiter\RateLimiterInterface::check()` to
`RateLimiterFactoryInterface` is a BC break with cross-repo reach that this sweep did not account for.
`pimcore/openid-connect`'s controllers pass their limiter factory into that method, so once their property
was retyped to the interface, PHPStan failed against the **released** `studio-backend-bundle v2026.2.3`,
which still declares the concrete parameter:

> Parameter #1 `$rateLimiterFactory` of method `...RateLimiterInterface::check()` expects
> `Symfony\Component\RateLimiter\RateLimiterFactory`, `Symfony\Component\RateLimiter\RateLimiterFactoryInterface` given.

Concrete → interface parameter is fine; interface → concrete parameter is not. The analysis said "callers
are unaffected", which was only true for callers passing a concrete factory — not for callers whose own
property had just been retyped. Landing this would require studio-backend to be merged *and released*
before openid-connect could go green, i.e. a release-ordered sequence across repos.

**Not urgent either way:** Symfony 8.0 removes only the *autowiring aliases*. The `limiter.*` services keep
`RateLimiterFactory` as their class, and all but two of the platform's injection points are bound
explicitly in yaml, so nothing here breaks at the Symfony 8 bump.

The inventory and the reasoning are preserved in `pimcore/platform-version`
(`migration-analysis/EXEC-RATELIMITER.md`) for when this is re-planned with the release ordering handled.

PR #1996, opened 2026-08-14
[Bug]: Skip relation columns whose class definition no longer exists

jcPimcore:
## Summary
- Same defect as #1995 (base `2025.4`), ported to `2026.2`. Renaming a Data Object class only updates that class itself; other classes' relation fields keep the old name in their `classes` allow-list.
- `AdvancedColumnCollector::buildRelationFields()` looked up every configured relation class unconditionally (aside from the existing `isFolderPseudoClass()` skip), so a stale reference threw `NotFoundException` and crashed the whole `available-columns` / `available-columns-for-relation` request instead of just omitting that one class.
- `ColumnConfigurationForRelationService` already handles this on `2026.2` (`resolveConfigurationForManyToManyObjectRelation()` / `resolveConfigurationForAdvancedManyToManyObjectRelation()` both null-check `getByName()`) — only `AdvancedColumnCollector` had the gap here.

Fixes https://github.com/pimcore/platform-version/issues/171

Not a forward-merge of #1995: `2025.4` and `2026.2` are divergent lines (`git merge-base --is-ancestor` confirms `2025.4` is not an ancestor of `2026.2`), so the fix is reapplied directly here rather than merged.

## Test plan
- Added `testGetColumnConfigurationsSkipsRelationClassesThatNoLongerExist` alongside the existing folder-pseudo-class tests in `AdvancedColumnCollectorTest`.
- Ran the full `AdvancedColumnCollectorTest` suite (6 tests) via Codeception in an isolated container checkout: all pass with the fix.
- Verified the new test fails with the exact `NotFoundException` against the pre-fix code, and that the 5 pre-existing tests (including the folder-pseudo-class cases) are unaffected either way.

### DIFF
--- a/src/Grid/Column/Collector/DataObject/AdvancedColumnCollector.php
+++ b/src/Grid/Column/Collector/DataObject/AdvancedColumnCollector.php
@@ -15,6 +15,7 @@ namespace Pimcore\Bundle\StudioBackendBundle\Grid\Column\Collector\DataObject;
 
 use Pimcore\Bundle\StaticResolverBundle\Models\DataObject\Objectbrick\DefinitionResolverInterface;
 use Pimcore\Bundle\StudioBackendBundle\Class\Repository\ClassDefinitionRepositoryInterface;
+use Pimcore\Bundle\StudioBackendBundle\Exception\Api\NotFoundException;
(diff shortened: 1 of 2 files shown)


jcPimcore:
obsolete.

Issue #1755, opened 2026-04-13
[Bug]: Path formatter broken with multiple localized field areas


ascheider:
### Studio API version

latest

### Steps to reproduce

Create two classes, for example category and product, i attach my example classes, you can reproduce the issue with it

[class_Product_export.json](https://github.com/user-attachments/files/26677313/class_Product_export.json)
[class_Category_export.json](https://github.com/user-attachments/files/26677314/class_Category_export.json)

Import the class definitions and create 2 categories and product.
Reference categories in product and save & publish product.

<img width="1117" height="280" alt="Image" src="https://github.com/user-attachments/assets/298f2dc3-0fb5-4339-a586-b32e3d9f0c19" />

After reload, the endpoint format-path throws the exception

<img width="1237" height="338" alt="Image" src="https://github.com/user-attachments/assets/79b42649-d80c-4c31-9574-8bac3c72efb2" />

<img width="1314" height="442" alt="Image" src="https://github.com/user-attachments/assets/ad883978-d81e-45c8-82c3-88fe1f24a428" />

The problem appear, because the $localizedfields->getChildren() is from the first localized fields area where name field is placed 

<img width="1034" height="514" alt="Image" src="https://github.com/user-attachments/assets/8d6e9520-cb8b-4e7f-a217-d2d500621c9f" />

I think the backend bundle need to resolve and merge all the localized fields and then check for it.

### Actual Behavior

The endpoint pimcore-studio/api/data-objects/format-path cause 500 error if multiple localized areas are in the class, the field is not resolved correctly

### Expected Behavior

The endpoint pimcore-studio/api/data-objects/format-path resolve the field correctly if multiple localized field areas present

martineiber:
Thank you for reporting this Issue
Fixed with https://github.com/pimcore/studio-backend-bundle/pull/1756

PR #2000, opened 2026-08-18
Fix BlockResolver failing on blocks containing layout components

kingjia90:
Resolves pimcore/platform-version#249

Blocks can contain layout components (`Fieldcontainer`, `Panel`, `Tabpanel`, …) next to data components. `BlockResolver::resolve()` filtered the block's raw `getChildren()` with a closure typed against `ClassDefinition\Data`, so as soon as a block contained any layout component the dot notation parser blew up with:

```
Uncaught PHP Exception TypeError: "…BlockResolver::{closure}(): Argument #1 ($field) must be of type
Pimcore\Model\DataObject\ClassDefinition\Data, Pimcore\Model\DataObject\ClassDefinition\Layout\Fieldcontainer given"
at BlockResolver.php line 66
```

This surfaces e.g. when selecting a value for a (dynamic) select option inside such a block in Studio UI, since `SelectOptionsService::getFieldDefinition()` resolves the field through the same parser — a `TypeError` is not an `Exception`, so it escapes the surrounding `catch` and ends up as a 500.

The child is now resolved via `Block::getFieldDefinition()`, which flattens layout components recursively — the same core API `FieldCollectionResolver` and `ObjectBrickResolver` already use. Lookups that previously worked are unaffected; only the flattening is new.

## Additional info
Added `tests/Unit/FieldDefinition/Parser/Resolver/BlockResolverTest.php` covering direct children, children behind one and multiple nested layout components, localized fields behind a layout component, and both `ParseException` paths. The unknown-child test reproduces the reported `TypeError` on the unpatched resolver.

Verified locally on PHP 8.4: full `Unit` suite green (505 tests) and PHPStan level 6 clean.

### DIFF
--- a/src/FieldDefinition/Parser/Resolver/BlockResolver.php
+++ b/src/FieldDefinition/Parser/Resolver/BlockResolver.php
@@ -63,12 +63,11 @@ final class BlockResolver extends AbstractResolver
             throw new ParseException('Class Definition has to be of type Block');
         }
 
-        $item = array_filter($fd->getChildren(), function (Data $field) use ($dotNotationParts) {
-            return $field->getName() === $dotNotationParts[1];
-        });
-        $item = reset($item);
+        // Blocks can contain layout components, therefore the children have to be flattened
+        // before looking up the data component by name
+        $item = $fd->getFieldDefinition($dotNotationParts[1]);
 
-        if (!$item) {
+        if ($item === null) {
             throw new ParseException(sprintf('Block field definition "%s" does not exist', $dotNotationParts[1]));
         }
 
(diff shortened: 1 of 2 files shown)


kingjia90:
Some history for context — this is the second half of an already-accepted fix, not a new approach.

The typed closure comes from #1083, where the line was originally `array_find($fd->getChildren(), function (Data $field) ...)`. `array_find()` is PHP 8.4+ and the bundle still supported 8.3 back then, so two "Replace array_find" commits in that same PR rewrote it to `array_filter()` + `reset()` and carried the `Data` hint over verbatim. The hint described the callback shape; it was never a claim that block children are always data components.

`LocalizedFieldService::getFieldDefinition()` was written in the same commit with the identical `getChildren()` + `Data`-closure shape, and was fixed in #1756 (resolving #1755) by switching to `getFieldDefinitions()` — same root cause, raw children are not the flattened field definitions. That fix only touched the localized-field service; `BlockResolver` kept the old pattern. This PR applies the same resolution to the remaining call site.

PR #2003, opened 2026-08-18
[Bug]: Parse notification creationDate in the application timezone

kingjia90:
Fixes pimcore/platform-version#335

### Problem

`Pimcore\Model\Notification\Dao::save()` writes `creationDate` / `modificationDate` as a naive wall-clock string via `date('Y-m-d H:i:s')`, i.e. in PHP's default timezone (set at boot from `general.timezone` in `Pimcore\Bootstrap`). No offset is stored alongside it.

`NotificationHydrator` read that same string back with `new Carbon($notification->getCreationDate(), 'UTC')` — assuming it was already UTC. It isn't; it's local time. With a non-UTC `general.timezone` the resulting epoch is shifted by the zone offset, so Studio's notification panel disagrees with the Application Logger and with the raw DB value for the same event:

```
notifications table: id=221189, creationDate = 2026-07-29 09:57:48   (correct local wall-clock time)
Application Logger (same event):               29.07.26, 09:57:48    (correct)
Studio notification panel (same event):        29.07.26, 11:57:48    (wrong, +2h)
```

All three hydrate methods were affected, so the list, the Mercure-published minimal payload and the detail view were all shifted.

For contrast, `ApplicationLoggerDb` converts to UTC explicitly on write and `LogHydrator` re-parses as UTC on read — a consistent round-trip. `Notification\Dao` never does the write-side conversion, only the hydrator made the (wrong) read-side UTC assumption.

### Fix

Parse the stored string with `date_default_timezone_get()` — the timezone it was actually written in — and collapse the three duplicated inline `Carbon` calls into a single private helper.

Read-side fix, as suggested in the issue: non-breaking and no migration needed, since existing rows are already naive local-time strings.

### Regression test

`tests/Unit/Notification/Hydrator/NotificationHydratorTest.php` covers:

- DST (`Europe/Berlin`, July) — all three hydrate methods
- outside DST (`Europe/Berlin`, January)
- UTC-configured instance — asserts the epoch is unchanged, so UTC installs see no behaviour change

Reverting just the timezone argument makes the test fail with exactly the deltas from the report: `1785319068` vs `1785311868` (+7200s, DST) and `1768471068` vs `1768467468` (+3600s, non-DST).

### Verification

- Full `Unit` suite: `OK (514 tests, 1182 assertions)`
- PHPStan level 6: `No errors`

### DIFF
--- a/src/Notification/Hydrator/NotificationHydrator.php
+++ b/src/Notification/Hydrator/NotificationHydrator.php
@@ -19,6 +19,7 @@ use Pimcore\Bundle\StudioBackendBundle\Notification\Schema\NotificationListItem;
 use Pimcore\Bundle\StudioBackendBundle\Notification\Schema\NotificationMinimal;
 use Pimcore\Model\Notification as NotificationModel;
 use Pimcore\Model\User;
+use function date_default_timezone_get;
 
 /**
  * @internal
@@ -33,7 +34,7 @@ final readonly class NotificationHydrator implements NotificationHydratorInterfa
             $notification->getTitle(),
             $notification->isRead(),
             (bool)$notification->getLinkedElementType(),
-            (new Carbon($notification->getCreationDate(), 'UTC'))->getTimeStamp(),
+            $this->getCreationTimestamp($notification),
             $this->getRecipientName($notification->getSender()),
         );
     }
@@ -45,7 +46,7 @@ final readonly class NotificationHydrator implements NotificationHydratorInterfa
             $notification->getType() ?? 'info',
             $notification->getTitle(),
             $notification->isRead(),
-            (new Carbon($notification->getCreationDate(), 'UTC'))->getTimeStamp(),
+            $this->getCreationTimestamp($notification),
             $notification->getRecipient()?->getId() ?? 0,
             $this->getRecipientName($notification->getSender()),
         );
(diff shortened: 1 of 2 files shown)


kingjia90:
Closing this — after digging into the history, **this is the wrong fix**, and I want to leave the reasoning here rather than silently drop it.

The hardcoded `'UTC'` in `NotificationHydrator` is **correct by design**. It isn't a Studio mistake — it was copied verbatim from the pre-Studio classic UI, `Pimcore\Model\Notification\Service\NotificationService::format()`, which has had the identical line since March 2023:

```php
$carbonTs = new Carbon($notification->getCreationDate(), 'UTC');
```

That line came from pimcore/pimcore#14741, which also shipped migration `Version20230321133700`, described as:

> Converts datetime/timestamp values to UTC for application logs and notifications.

Its `up()` converts `application_logs.timestamp`, `notifications.creationDate` **and** `notifications.modificationDate` from the local timezone to UTC, and it's documented in the upgrade notes. So UTC storage is the intended, migrated, documented convention for these columns — reading them as UTC is right.

What that PR did *not* do is update the writer. It fixed the readers and made `ApplicationLoggerDb` write UTC, but left `Pimcore\Model\Notification\Dao::save()` on `date('Y-m-d H:i:s')` (local time). `git log -S UTC -- models/Notification/Dao.php` returns nothing — that Dao has never written UTC since the feature landed in #3401.

**So the 2023 change was incomplete: it migrated the data and fixed the readers, and forgot the writer.** Since then the `notifications` table holds a mix — rows written before the migration are UTC and display correctly, everything written after is local wall-clock and displays shifted by the zone offset. That's the reported +2h.

Aligning the Studio reader with the buggy writer (what this PR does) would fix new rows in Studio but break the pre-migration rows that genuinely *are* UTC, leave the classic UI still wrong, make the two UIs disagree with each other, and entrench a violation of a convention that has a migration behind it.

The fix belongs in `pimcore/pimcore` on the write side, so that no reader anywhere needs changing and both UIs become correct at once. Follow-up PR incoming there; this one is superseded.

PR #2004, opened 2026-08-19
Normalize entry params

alexej-d:
Resolves #321 

## Additional info
Needed in addition to the [main patch](https://github.com/pimcore/studio-ui-bundle/pull/3946).

### DIFF
--- a/src/Email/Controller/SearchController.php
+++ b/src/Email/Controller/SearchController.php
@@ -1,0 +1,103 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * This source file is available under the terms of the
+ * Pimcore Open Core License (POCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (https://www.pimcore.com)
+ *  @license    Pimcore Open Core License (POCL)
+ */
+
+namespace Pimcore\Bundle\StudioBackendBundle\Email\Controller;
+
+use OpenApi\Attributes\Post;
+use Pimcore\Bundle\StudioBackendBundle\Controller\AbstractApiController;
+use Pimcore\Bundle\StudioBackendBundle\Email\Schema\EmailLogEntry;
+use Pimcore\Bundle\StudioBackendBundle\Filter\Attribute\Request\CollectionRequestBody;
+use Pimcore\Bundle\StudioBackendBundle\Gdpr\Schema\GdprDataRow;
+use Pimcore\Bundle\StudioBackendBundle\Gdpr\Service\GdprManagerServiceInterface;
+use Pimcore\Bundle\StudioBackendBundle\MappedParameter\CollectionFilterParameter;
+use Pimcore\Bundle\StudioBackendBundle\OpenApi\Attribute\Property\GenericCollection;
+use Pimcore\Bundle\StudioBackendBundle\OpenApi\Attribute\Response\Content\CollectionJson;
+use Pimcore\Bundle\StudioBackendBundle\OpenApi\Attribute\Response\DefaultResponses;
+use Pimcore\Bundle\StudioBackendBundle\OpenApi\Attribute\Response\SuccessResponse;
+use Pimcore\Bundle\StudioBackendBundle\OpenApi\Config\Tags;
+use Pimcore\Bundle\StudioBackendBundle\Util\Constant\HttpResponseCodes;
+use Pimcore\Bundle\StudioBackendBundle\Util\Constant\UserPermissions;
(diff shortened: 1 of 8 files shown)


jcPimcore (approved):
Good fix — `EmailLogEntryParameter`'s constructor requires `?string` under `strict_types=1`, but `Mail::setParam()` accepts `mixed`, so any logged email with a bool/int/array template param would `TypeError` on `GET /emails/{id}/params`. `normalizeParamValue()` fixes this at the right spot (the service producing the value, not the schema contract), and correctly leaves the resend path (`setEmailDocumentContent`) untouched since that needs the original typed value.

One ask before merging: this repo has a solid per-service unit-test convention (`tests/Unit/<Module>/Service/*ServiceTest.php`), but there's no `tests/Unit/Email` suite at all. Since this fixes an actual crash, could you add a small test for `getEntryParams()`/`normalizeParamValue` covering bool/int/array values?

alexej-d:
> Good fix — `EmailLogEntryParameter`'s constructor requires `?string` under `strict_types=1`, but `Mail::setParam()` accepts `mixed`, so any logged email with a bool/int/array template param would `TypeError` on `GET /emails/{id}/params`. `normalizeParamValue()` fixes this at the right spot (the service producing the value, not the schema contract), and correctly leaves the resend path (`setEmailDocumentContent`) untouched since that needs the original typed value.
> 
> One ask before merging: this repo has a solid per-service unit-test convention (`tests/Unit/<Module>/Service/*ServiceTest.php`), but there's no `tests/Unit/Email` suite at all. Since this fixes an actual crash, could you add a small test for `getEntryParams()`/`normalizeParamValue` covering bool/int/array values?

Check the new commit @jcPimcore

jcPimcore:
Confirmed — `EmailLogServiceTest::testGetEntryParamsNormalizesNonStringValues` correctly exercises the fixed path (bool → `'true'`, int → `'118'`, array → JSON) and matches this repo's Codeception/`makeEmpty` convention. Thanks!

jcPimcore:
Hi @alexej-d 
Thank you for the update! Looks good! 👍 
Since it is related to https://github.com/pimcore/studio-ui-bundle/pull/3946 we will not merge this immediately but waiting for the other PR.

alexej-d:
> Hi @alexej-d Thank you for the update! Looks good! 👍 Since it is related to [pimcore/studio-ui-bundle#3946](https://github.com/pimcore/studio-ui-bundle/pull/3946) we will not merge this immediately but waiting for the other PR.

I have adjusted my code in both bundles @jcPimcore

jcPimcore:
Hi @alexej-d 
Great work again! 👍 

There was something found which was not adressed before. Can you please implement this? 


## Security: /emails/search broadens exposure to an ORDER-BY injection sink

POST /emails/search (added in studio-backend-bundle#2004) passes the request's sortFilter.key straight through EmailLogRepository::applySearchOptions() → Listing::setOrderKey() → AbstractDao::getOrder(), which builds ORDER BY <key> <direction> via raw string concatenation — no allow-list, no escaping.

This sink already exists (reachable today via POST /gdpr/search?provider=sent_mails), but that path requires the gdpr_data_extractor permission. The new endpoint gates the same sink behind emails instead — a much more commonly granted permission — so this PR meaningfully widens who can reach it.

Suggested fix: allow-list the sortable columns at EmailLogRepository::applySearchOptions() before calling setOrderKey(), e.g.:

private const array SORTABLE_KEYS = ['id', 'sentDate', 'from', 'to', 'subject'];

$sortFilter = $filter->getSortFilter();
if ($sortFilter->getKey() && in_array($sortFilter->getKey(), self::SORTABLE_KEYS, true) && $sortFilter->getDirection()) {
    $listing->setOrderKey($sortFilter->getKey());
    ...
}

falling back to the existing default order otherwise. That's the smallest fix at the right boundary — it doesn't touch shared core (AbstractDao) and mirrors how a fixed version of this pattern should look.

---

Thanks again in advance!

Jochen

jcPimcore (changes requested):
Security fix as commented.

alexej-d:
> Security fix as commented.

Hi @jcPimcore – the fix is implemented as requested. I also added a test

jcPimcore:
Hi @alexej-d 
Thanks again for the fast answer and the committed fix!
Everything is green now.

I will do I final functional check. If everything works as expected, we can merge this.
Will inform you soon about it 👍 

Thx and kind regards,

Jochen

jcPimcore (approved):

jcPimcore:
Hi @alexej-d 
Tested and merged, too. Will also be part of the 2026.3 release.
Thanks a lot for this great contribution!

Kind regards,
Jochen

PR #2006, opened 2026-08-20
Fix undefined array key error_pages in SystemSettingsProvider

kingjia90:
## Summary

`SystemSettingsProvider::getDocumentSettings()` accessed `documents['error_pages']['default']` and `documents['error_pages']['localized']` directly, without checking whether those keys exist. When the system settings config does not (yet) contain a fully populated `error_pages` structure — e.g. right after switching `config_location.system_settings` to `settings-store` before it has ever been saved — this raises `Undefined array key "error_pages"` / `"localized"` warnings. Under Symfony's debug error handler these warnings are promoted to fatal `ErrorException`s, so the Studio "Appearance & Branding" settings screen (and any other caller of `getSettings()`) crashes with a 500.

Fixes #206.

## Changes

- `getDocumentSettings()` now safely defaults a missing `error_pages` (and `error_pages.localized`) to an empty array instead of relying on the keys always being present.
- Added `SystemSettingsProviderTest` with two regression tests that reproduce the exact warnings from the issue (verified they fail against the pre-fix code at the exact reported lines) and pass after the fix.

## Test plan

- [x] `vendor/bin/codecept run Unit` — 459 tests, all green
- [x] `vendor/bin/phpstan analyse` — no errors
- [x] `vendor/bin/php-cs-fixer fix --dry-run` — no style violations
- [x] New regression tests fail on the unmodified code with the same `Undefined array key` warnings reported in the issue, and pass after the fix

### DIFF
--- a/src/Setting/Provider/SystemSettingsProvider.php
+++ b/src/Setting/Provider/SystemSettingsProvider.php
@@ -132,22 +132,26 @@ final readonly class SystemSettingsProvider implements SettingsProviderInterface
             return [];
         }
 
-        $errorPage = $this->getErrorDocument($documents['error_pages']['default']);
+        $errorPages = $documents['error_pages'] ?? [];
+
+        $errorPage = $this->getErrorDocument($errorPages['default'] ?? null);
         if ($errorPage) {
-            $documents['error_pages']['default'] = $this->elementDataService->getRelatedElementData(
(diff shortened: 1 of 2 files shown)


kingjia90:
Closing in favor of #2007, which fixes the same `SystemSettingsProvider` "Undefined array key `error_pages`" crash and now references this PR's linked issue (#206) as well.

PR #2011, opened 2026-08-20
Add paginated object-dependencies endpoint for users, avoid OOM

jcPimcore:
Resolves: https://github.com/pimcore/platform-version/issues/308 Studio UI
Resolves: https://github.com/pimcore/admin-ui-classic-bundle/issues/1106 Studio UI

`GET /user/{id}` OOMs / times out when the user is referenced by a very large number of DataObjects via a `User`-type class field, because `ObjectDependenciesService::getDependenciesForUser()` unconditionally hydrated every matching object and embedded the full, unbounded list in every response (same class of bug as pimcore/admin-ui-classic-bundle#1106, just not yet reported for Studio).

- Add `ObjectDependenciesRepository`/`Interface`: reproduces the class/field discovery, but counts matches per class first (cheap, non-hydrating) and only ever hydrates the slice of objects that falls inside a requested `offset`/`limit` window - bounded regardless of total match count or page depth.
- Add a new paginated `GET /user/{id}/object-dependencies` endpoint (`GetObjectDependenciesController`), modeled on the existing generic element Dependencies tab's `CollectionController`, to browse the full list. Enforces a `MAX_PAGE_SIZE` of 100 (matching the largest page size selectable in the Studio UI) before offset/limit are computed, and resolves/authorizes the target user first (404 if unknown, 403 if it's an admin and the caller isn't) instead of silently returning 200 with an empty collection.
- **No schema-breaking change, but a real behavioral impact worth knowing about**: `objectDependencies` on the main `User` schema keeps its existing `dependencies`/`hasHidden` fields exactly as they were (same names, same types) - it only gains a new field, `totalItems`, alongside them. `totalItems` is marked *optional* in the OpenAPI schema (even though it's always present in the actual response) specifically because `studio-ui-bundle` is an actively-published npm package and this type is re-exported through its public SDK entry point - making it required would generate a non-optional TS field and break any existing consumer code constructing/mocking a value of that type. However, `dependencies` previously contained the user's *complete* list of referencing objects, and now stops at 20. Any existing integration that assumed `dependencies` was exhaustive (rather than checking `totalItems`) will now silently see only the first 20 - such integrations must start reading `totalItems` and, if it exceeds `dependencies.length`, call the new paginated endpoint for the rest. `hasHidden` is affected the same way: it now only reflects permission-denied objects within that 20-item window, not across the full list as before. `getPreviewForUser()` is a new service method used only by `UserHydrator` for this embedded preview, kept separate from `getPaginatedDependenciesForUser()` (used by the new endpoint, returns the generic `Collection` type).
- Multi-`User`-field classes: dependency matching now joins conditions with `OR` instead of `AND` - an object referencing the user via any one of several `User`-type fields was previously excluded from both the preview and the endpoint unless *every* field matched.
- Deterministic `ORDER BY id` on the underlying listings, so offset/limit paging is stable across requests (verified: two adjacent pages against real data have zero overlap and no gap).
- Upgrade note added/refined across a few passes to accurately describe the above.

## Additional info
Verified against a real 50,000-object repro (a backend user referenced by 50,000 `DataObject`s via a `user` field, via the real DI container against the live DB, not just unit-mocked):
- The new endpoint returns exactly the requested page (tested page 1, a mid page, and the true last page) with correct `totalItems`; memory stays flat across page depth (no growth as offset increases).
- `UserHydrator`'s embedded preview returns `class=ObjectDependencies, hasHidden=false, dependencies=20, totalItems=50000` in ~0.2s / ~90MB peak (bounded, no OOM).
- `Schema/User.php` diffs as byte-for-byte identical to its pre-change state; `ObjectDependencies.php` diffs as purely additive (only `totalItems` + its getter added, `dependencies`/`hasHidden` untouched).
- A focused unit test (`ObjectDependenciesRepositoryTest`) covers the cross-class pagination windowing math directly (class-boundary pages, deep offsets, exhausted budget, exact boundaries), since the surrounding dynamic class-discovery/listing code has no fixtures in this bundle's unit suite.

Related: pimcore/studio-ui-bundle's `fix/308-paginated-user-object-dependencies` (frontend for this endpoint), pimcore/ee-admin-ui-classic-bundle's `fix/308-cap-user-object-dependencies` (short-term cap for the classic admin UI, same root issue).


jcPimcore:
Addressed all 5 points from the Copilot review in 2ca20d3d:

- `ObjectDependenciesRepository`: multi-field `User` conditions now joined with `OR` instead of `AND` (an object referencing the user via any one of several `User`-type fields was previously excluded unless *every* field matched).
- Added a deterministic `ORDER BY id` so offset/limit paging is stable across requests — verified against the real 50k-row repro data: two adjacent pages now have zero overlap and no gap.
- `GetObjectDependenciesController` now resolves and authorizes the target user (404 if missing, 403 if it's an admin and the caller isn't, mirroring `UserService::getUserById()`) before querying dependencies, instead of silently returning 200 with an empty collection for an unknown/unauthorized id.
- Added a `MAX_PAGE_SIZE = 100` cap (matching the largest selectable page size in the Studio UI), enforced before offset/limit are computed, since `CollectionParameters` has no upper bound on its own and an unbounded `pageSize` would reintroduce the OOM this endpoint exists to prevent.
- Extracted the cross-class windowing arithmetic into a pure `resolveClassWindow()` method and added `ObjectDependenciesRepositoryTest` covering class-boundary pages, deep offsets, exhausted budget, and exact boundaries — the surrounding dynamic class-discovery/listing code still has no unit fixtures in this bundle's suite, but the math that determines pagination correctness is now directly tested.

jcPimcore:
Addressed both points from this review in d8d01a84:

- Fixed the route typo: the upgrade note said `GET /users/{id}` (plural); the actual route is `GET /user/{id}` (singular, `GetUserController`).
- Reworded the "not a breaking change" note - it was an overclaim. To be precise: the `objectDependencies` **schema** genuinely has no breaking change (`dependencies`/`hasHidden` keep their names, types, and meaning; `totalItems` is purely additive). But you're right that there's a real **behavioral compatibility impact**: `dependencies` previously contained the complete list and now caps at 20. Any consumer that assumed completeness rather than checking `totalItems` will now silently see a partial view. The note states that plainly now instead of a blanket "not breaking" claim. Also updated the PR description to match.

src/User/Controller/GetObjectDependenciesController.php
```diff
@@ -0,0 +1,119 @@
...
+    ): JsonResponse {
+        if ($parameters->getPageSize() > self::MAX_PAGE_SIZE) {
+            throw new InvalidFilterException(sprintf('pageSize must not exceed %d', self::MAX_PAGE_SIZE));
+        }
+
+        $targetUser = $this->userRepository->getUserById($id);
+
+        if ($targetUser->isAdmin() && !$this->securityService->getCurrentUser()->isAdmin()) {
+            throw new ForbiddenException('Only admins can view other admins');
+        }
```
martineiber: No Logic in the controller!

src/User/Repository/ObjectDependenciesRepository.php
```diff
@@ -0,0 +1,144 @@
...
+        $matchesBeforeCurrentClass = 0;
+
+        foreach ($this->getClassesWithUserField() as $className => $fields) {
+            $list = $this->createListingForClass($className, $fields, $userId);
+            $classMatchCount = $list->count();
+            $totalItems += $classMatchCount;
+
+            $stillNeeded = $limit - count($items);
+            if ($stillNeeded > 0) {
+                $window = $this->resolveClassWindow($matchesBeforeCurrentClass, $classMatchCount, $offset, $limit, $stillNeeded);
```
martineiber: Fix Sonar findings.

src/User/Repository/ObjectDependenciesRepository.php
```diff
@@ -0,0 +1,144 @@
...
+                if ($window !== null) {
+                    [$localOffset, $localLimit] = $window;
+                    $list->setOffset($localOffset);
+                    $list->setLimit($localLimit);
+                    // A per-class Listing (e.g. Issue1106\Listing) only ever loads Concrete
+                    // instances of that class, but Listing::load() is typed to the broader
+                    // DataObject; filter to narrow it back for callers that expect Concrete.
+                    $loaded = array_filter(
+                        $list->load(),
+                        static fn (\Pimcore\Model\DataObject $object): bool => $object instanceof Concrete
```
martineiber: `use Pimcore\Model\DataObject;`
```suggestion
                        static fn (DataObject $object): bool => $object instanceof Concrete
```

src/User/Schema/ObjectDependencies.php
```diff
@@ -22,7 +22,10 @@
  */
 #[Schema(
     title: 'User Object Dependencies',
-    description: 'User Object Dependencies',
+    description: 'User Object Dependencies. dependencies is a bounded preview - use '
```
martineiber: Schema description should have max 100 chars.

martineiber (changes requested):

src/User/Repository/ObjectDependenciesRepository.php
jcPimcore: done

martineiber (approved):

PR #2017, opened 2026-08-27
Fix: stale Perspective no longer restricts Superadmin (2025.4)

robertSt7:
Cherry-pick of #1971 onto `2025.4`.

Related issue: https://github.com/pimcore/studio-ui-bundle/issues/3615 (reported on 2025.4)

## Why

`#1971` fixed this on `2026.2` only. `2025.4` is still affected and is actively maintained, so it never received the fix — and `2025.4` forward-merges into `2026.2`, so landing it here keeps the lines consistent instead of divergent.

## What

Cherry-picked verbatim (`6d00c8fa`), no adaptation needed — the commit applied cleanly and the resulting `UserPerspectiveService.php` is **byte-identical to `2026.2`'s** (blob `b7820485`).

- `getAllUserPerspectives()` short-circuits to an empty list for admins; the existing "empty means unrestricted" fallback in all three call sites handles the rest.
- `getActivePerspective()` returns the default perspective for admins, so a stale stored active value cannot keep them on the old restrictive view.

## Test

Brings across `tests/Unit/User/Service/UserPerspectiveServiceTest.php` (143 lines) from `#1971`. The file did not exist on `2025.4`, so it lands as a clean add.

## Verified / not verified

- Cherry-pick applied with **no conflicts**; diff is +151 lines across 2 files, nothing beyond the upstream commit.
- `php -l` clean on both files.
- **Not** run locally: Codeception and PHPStan — this branch has no `vendor/`. Relying on CI.

## Note for reviewers

`getActivePerspective()`'s unconditional admin short-circuit discards a perspective an admin explicitly selected (it is re-read on every application load), so an admin's own choice does not persist. That behaviour is **already live on `2026.2`** via `#1971`; this PR deliberately does not diverge from it. Worth addressing on both lines together in a follow-up rather than only here.

### DIFF
--- a/src/User/Service/UserPerspectiveService.php
+++ b/src/User/Service/UserPerspectiveService.php
@@ -79,6 +79,10 @@ final readonly class UserPerspectiveService implements UserPerspectiveServiceInt
 
     public function getActivePerspective(UserInterface $user): string
     {
+        if ($user->isAdmin()) {
+            return Perspectives::DEFAULT_ID->value;
+        }
+
         $activePerspective = $this->repository->getUserActivePerspective($user->getId());
         $userPerspectives = $this->getAllUserPerspectives($user);
         if (empty($userPerspectives)) {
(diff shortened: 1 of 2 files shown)


robertSt7:
@copilot review

PR #2019, opened 2026-08-31
Allow patching a multiselect to empty

xIrusux:
Because clearing a multiselect through a merge is a legal operation the core adapter's type signature forbade.

**What the merge sent:** object 333's `color` is empty; you chose to pull that side over, so the merger saved `color: { action: 'replace', data: null }` — "replace the target's value with nothing", i.e. clear the field. That's the correct payload for this merge, not a frontend bug.

**Where it exploded:** `MultiSelectAdapter::handlePatch` handles `replace` by returning the payload's `data` verbatim — but its return type was declared `array`, so `data: null` throws a `TypeError` and the whole PATCH 500s (MultiSelectAdapter.php:66).

**Why widening to `?array` is the right fix, not a workaround:** the caller already supports null — `PatchService::patchEditableData` passes the adapter's return straight into `$element->setValue($key, $value)` with no null check, and `setValue(key, null)` is Pimcore's normal "clear this field". The add/remove paths in the same adapter also cope with null existing values. Only the `replace` path's signature was narrower than its own contract. One character (`?`) restores it.

### Verified

- Reproduced via the object merger (embedded in the backend-power-tools compare & merge wizard): merging an object with an empty multiselect over one with values 500ed; with this change the PATCH clears the field.
- Regression test added (`MultiSelectAdapterTest`, following the `NumericRangeAdapterTest` pattern): replace with `data: null` returns null, replace with values returns them, non-array input returns null. 3/3 green locally via `vendor/bin/codecept run Unit`.

### DIFF
--- a/src/DataObject/Data/Adapter/MultiSelectAdapter.php
+++ b/src/DataObject/Data/Adapter/MultiSelectAdapter.php
@@ -59,7 +59,7 @@ final readonly class MultiSelectAdapter implements SetterDataInterface
         string $key,
         array $newData,
         ?FieldContextData $contextData = null
-    ): array {
+    ): ?array {
         $action = $newData[PatchDataKeys::ACTION->value];
         $fieldData = $newData[PatchDataKeys::DATA->value];
         if ($action === null || $action === PatcherActions::REPLACE->value) {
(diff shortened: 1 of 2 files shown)


xIrusux:
Superseded by the 2025.4-based PR — the fix belongs on the lowest affected release line and forward-merges up.

xIrusux:
Superseded by #2020.

PR #2022, opened 2026-08-31
[OAuth] Multi-application OAuth: public contracts, audience binding, scope catalogue

fashxp:

Makes the embedded OAuth 2.1 server usable by more than one application, and closes the gap that created: with a second protected resource in play, an unbound token would have opened all of them.

### 1. Public contracts

Drops `@internal` from the types a bundle needs to act as a *resource server*, and adds two for scopes.

| Type | Why it is public |
|---|---|
| `TokenValidatorInterface` | validate a bearer token without duplicating signature, expiry, audience and revocation checks |
| `ResolvedAccess` | its return type |
| `ResourceRegistryInterface` | register an endpoint as a protected resource, which is what makes its RFC 9728 metadata resolvable |
| `ProtectedResource` / `ProtectedResourceMetadata` | argument and return types of the above |
| `ScopeProviderInterface` / `ScopeRegistryInterface` | contribute your own scope namespace |

`CanonicalUri` stays internal: the registry and the validator canonicalise on both sides, so callers never need it.

### 2. Audience binding (RFC 8707), with `resource` required

A client names the resource it wants a token for; the server validates it against the registry, refuses an unknown one, stamps the **canonical** form as `aud`, and `TokenValidatorInterface::validate()` refuses a token whose audience does not name the resource it is presented at. The binding survives a refresh.

**`resource` is mandatory.** An authorization request that omits it is refused with `invalid_request`. The alternative — accepting an audience-less token everywhere — leaves exactly the cross-application hole this PR exists to close, and there is no OAuth surface in the wild yet whose compatibility is worth preserving.

The practical consequence is that clients which do not implement RFC 8707 stop working, and that is deliberate. Two were found and fixed on the application side rather than by relaxing the rule: Swagger UI (passes the resource via `additionalQueryStringParams`) and MCP Inspector, which sends `resource` only when it can discover protected-resource metadata from the endpoint it was given.

**Why now rather than later.** It was deferred on the reasoning that per-application admission covers access control. That holds *within* an application; it does not hold *between* them. Once `data-hub-simple-rest` became a second resource, a token consented to for one application authenticated against the other as the same Pimcore user. Adding the second resource is what made the pre-existing gap exploitable.

### 3. The binding is carried in the token record, not in a forked league

An earlier revision mirrored two `league/oauth2-server` internals to move the resource through the flow. Both are gone. A nullable `resource` column on the token record carries it instead:

- `src/OAuth/Server/ResponseType/ResourceBearerTokenResponse.php` is **deleted**.
- `LoopbackAuthCodeGrant::completeAuthorizationRequest()` is reduced to a six-line hook that captures the resource and delegates to `parent::`, rather than reproducing league's body.

The commit deletes 185 lines against 282 added (the additions are mostly the migration, the record plumbing and tests). The point is not the count but that no upstream body is reproduced any more, so the flow no longer breaks silently when league changes internally.

### 4. Scope catalogue

Scopes were hardcoded in four places. They now come from tagged `ScopeProviderInterface` services, so a bundle ships its own namespace (`datahub:read`) instead of reusing `mcp:*`, and discovery advertises the union. A dynamic registration that omits `scope` now yields no scope rather than "whichever provider loaded first", which depended on bundle registration order.

### 5. A functional break this surfaced

Enforcing `aud` exposed a mismatch that was invisible while nothing compared it: MCP resources register as `<issuer>/pimcore-mcp/studio/<slug>`, but the authenticator validated against `<host>/pimcore-mcp`. A client that discovered a server and correctly requested a token for it was then refused **at that very server**. The authenticator now derives the per-server URI the same way registration does. Verified live: a token bound to `…/studio/<slug>` authenticates there, and the same token is refused at a different resource.

### 6. Documentation

New **OAuth-Protected Applications** page: the authorization-server / resource-server split, the public contracts, the anatomy of an application, and a blueprint covering both integration shapes actually in use (a firewall, as the MCP servers do; an existing request pipeline, as Data Hub Simple REST does). The surrounding pages are reframed so OAuth reads as a platform capability rather than an MCP feature, and MCP server management states it is a configuration topic independent of how a caller authenticated.

Two corrections where docs promised more than the code delivered: the MCP page claimed the authenticator "validates the token's signature and audience" when it did not, and resource URIs were described as an isolation boundary before they were one. The new page has an explicit "what the platform does not do yet" section, which still names scope enforcement.

## Known gaps, stated plainly

- **Scopes are advertised and consented to but never enforced at call time.** A token with no scope behaves like a fully scoped one. Documented; not fixed here.
- The consent screen does not yet show which resource is being authorised, and renders a scope it does not recognise (`datahub:read`) as the raw string, because the label map in `studio-ui-bundle` lists only `mcp:*`.

## Verification

Unit suite green: **871 tests, 1967 assertions**, including new coverage for resource validation, canonicalisation, and a request that names no resource.

Manually end-to-end against a running stack: the real authorization-code + PKCE flow, a token carrying `aud`, accepted at its own resource and refused at another; an authorization request without `resource` refused with `invalid_request`; and the binding intact across a refresh. Discovery verified through the 401 challenge and the metadata document, for the Studio MCP servers and for both Data Hub Simple REST surfaces.

php-cs-fixer clean on the changed paths. PHPStan not run locally (no dev dependencies in the parent project); note the `static-analysis` failure on this PR is at `DependencyInjection/Configuration.php:867`, a file this branch does not touch, inherited from the base branch.


fashxp:
replaced by https://github.com/pimcore/studio-backend-bundle/pull/2028 and https://github.com/pimcore/studio-backend-bundle/pull/2029

PR #2026, opened 2026-09-03
Fix falsy value check dropping 0/false in field collection sub-fields

jcPimcore:
## Summary
- `FieldCollectionsAdapter::processCollectionRaw()` used `!$elementValue` to decide whether a field-collection sub-field had a submitted value.
- That's `true` for `0`, `false`, `""`, `"0"` in PHP, so any such legitimate value submitted for a field-collection sub-field (including a pre-filled numeric default of `0`) was silently dropped before it ever reached the field's own setter adapter.
- Changed the guard to `$elementValue === null`, so a genuinely absent field is still skipped, but falsy-but-real values are kept.

## Test plan
- Added `tests/Unit/DataObject/Data/Adapter/FieldCollectionsAdapterTest.php`, calling `processCollectionRaw()` via reflection with a real `Fieldcollection\Definition` registered in `RuntimeCache`.
- Ran the suite locally: all 4 new tests pass against the fix, and I confirmed 3 of them fail against the pre-fix code (temporarily reverted the fix, reran, confirmed failures, then restored it) — verified regression test.

Refs PEES-1279
Resolves https://github.com/pimcore/service-operations/issues/953

### DIFF
--- a/src/DataObject/Data/Adapter/FieldCollectionsAdapter.php
+++ b/src/DataObject/Data/Adapter/FieldCollectionsAdapter.php
@@ -166,9 +166,8 @@ final readonly class FieldCollectionsAdapter implements
         $collectionDef = Fieldcollection\Definition::getByKey($collectionRaw['type']);
 
         foreach ($collectionDef?->getFieldDefinitions() as $elementName => $fd) {
-            $elementValue = $blockElement[$elementName] ?? null;
             $adapter = $this->dataAdapterService->tryDataAdapter($fd->getFieldType());
-            if ($elementValue === null || !$adapter) {
+            if (!$adapter || !is_array($blockElement) || !array_key_exists($elementName, $blockElement)) {
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


jcPimcore:
Good catch — fixed in ebfc3e83:

- `processCollectionRaw()` now checks `array_key_exists($elementName, $blockElement)` instead of testing the value for `null`, matching the existing pattern in `ObjectBricksAdapter::getCollectionData()`. A present `null` now reaches the setter adapter (so a field-collection sub-field can be cleared on non-patch saves), while a genuinely absent key is still skipped.
- Added a guard for `$blockElement` not being an array at all (no `data` key on the collection item), so that case doesn't get misread as "every field explicitly null".
- Added regression tests for both: explicit-null-passthrough, and a collection item with no `data` key.
- Ran the full `tests/Unit/DataObject/` suite locally — 43 tests, 93 assertions, no regressions.

PR #2047, opened 2026-09-16
[OAuth] Give a request without scope the resource's declared scopes

fashxp:
Stacked on #1958.

### Problem

An authorization request that omits `scope` passed through as an empty set. The consent screen then said
"No specific permissions were requested", and the resulting token still reached the resource, because nothing
checks scopes at call time. Found through Datahub Simple REST (pimcore/data-hub-simple-rest#356), which now
refuses such a token itself; Studio's own MCP servers had the same gap.

### Change

RFC 6749 section 3.3 lets the server either apply a default or refuse. Refusing would break clients that work
today, so `LoopbackAuthCodeGrant::narrowToResource()` now defaults an empty request to every scope the resource
declares. The consent screen shows those scopes and the token carries them. A resource declaring no scopes is
unchanged.

The docs for protected applications now tell them to check their scope even when every operation has the same
privilege. The reason is refresh: a client may ask for fewer scopes than were granted, including none, and no
consent screen is shown then.

### Verified

- Unit: 1039 tests pass, PHPStan clean. The new grant test and the new end-to-end case in
  `ResourceBindingLifecycleTest` both fail if the default is removed.
- Live, request without `scope`:
  - Datahub MCP: consent `['datahub:read']`, token `scope: datahub:read`, MCP 200.
  - Studio MCP: consent `['mcp:read', 'mcp:write']`, token `scope: mcp:read mcp:write`.


fashxp:
Closing. The default was deliberately left empty: RFC 6749 §3.3 allows it, it doesn't break clients that omit `scope`, and it's the least-privilege choice once resources enforce their scopes (Datahub does now, Studio MCP will with per-tool scopes). Defaulting to every declared scope would hand such a client maximal privilege. The misleading consent wording for an empty scope list is handled in the UI instead, and the "check your scope" guidance for protected applications moves into #1958.

